PR #75, opened 2023-01-19
[Operator] Add reduce_f16 and squeeze as Reduce's resolve variants

hjjq:
The original reduce schedule is not optimized for FP16 input.
1) Added an FP16-optimized reduce schedule.
2) Simplified cases where size of all reduce dimensions are 1 to squeeze.
3) Added resolve rules for 1) and 2). Move relevant files into a single directory.
4) Added tests for FP16 reduce. (Unrelated: Added an extra common test case in conv2d)

### DIFF
--- a/python/hidet/graph/ops/definitions/reduce/reduce_f16.py
+++ b/python/hidet/graph/ops/definitions/reduce/reduce_f16.py
@@ -11,16 +11,11 @@
 # limitations under the License.
 from typing import List, Callable, Any, Union, Optional, Dict, Sequence
 from hidet.ir import IRModule
-from hidet.ir.expr import Expr
 from hidet.ir.primitives import active_mask, shfl_down_sync, shfl_sync
 from hidet.ir.compute import ReduceOperation, reduce
 from hidet.ir.layout import DataLayout, StridesLayout
-from hidet.ir.mapping import TaskMapping
 from hidet.ir.type import data_type, TensorType, TensorPointerType
-from hidet.graph.ops.definitions.matmul.batch_matmul import BatchMatmulOp
-from hidet.transforms.tools import fuse_and_pack
 from hidet.lang import f16, f32, i32, spatial, repeat, tensor, attr, grid, printf, cast, tensor_pointer
-from hidet.lang.mapping import repeat, spatial
 from hidet.lang.cuda import blockIdx, threadIdx, syncthreads, register_tensor
 from hidet.transforms.tools import add_packed_func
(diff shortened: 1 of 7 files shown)


python/hidet/graph/ops/definitions/reduce/reduce_f16.py
```diff
@@ -0,0 +1,298 @@
...
+from hidet.graph.ops.definitions.utils import (
+    Task,
+    Operator,
+    Tensor,
+    TensorNode,
+    compute,
+    input_like,
+    normalize_dim,
+    ReduceType,
+)
```
yaoyaoding, line 29: Consider to split the import into two lines.

python/hidet/graph/ops/definitions/reduce/reduce_f16.py
```diff
@@ -0,0 +1,298 @@
...
+from hidet.utils import prod
+
+
+class ReduceF16Task(Task):
+    def __init__(
+        self, x: TensorNode, dims: List[int], keep_dim: bool, reduce_type: ReduceType, accumulate_dtype: str = 'float32'
+    ):
+
```
yaoyaoding, line 37: raise an error if the data type of x is not float16.

python/hidet/graph/ops/definitions/reduce/reduce_f16.py
```diff
@@ -0,0 +1,298 @@
...
+                rv[0] = ro.combine(rv[0], shfl_down_sync(mask, rv[0], 16, 32))
+                rv[0] = ro.combine(rv[0], shfl_down_sync(mask, rv[0], 8, 32))
+                rv[0] = ro.combine(rv[0], shfl_down_sync(mask, rv[0], 4, 32))
+                rv[0] = ro.combine(rv[0], shfl_down_sync(mask, rv[0], 2, 32))
+                rv[0] = ro.combine(rv[0], shfl_down_sync(mask, rv[0], 1, 32))
+                rv[0] = shfl_sync(mask, rv[0], 0, 32)
+                rv[0] = ro.finalize(acc=rv[0], size=reduce_extent)
+                if threadIdx.x == 0:
+                    for indices in remain_layout.on(blockIdx.x):
+                        y.write(indices, rv[0], protected=True)
```
yaoyaoding, line 160: In this case, seems we do not need 'protected=True' as the number of thread blocks is exactly set to the required number of workers. Using protected write would add an condition statement to protect the write, one our bound analyzer can not remove it for us, there might be some performance degragation. 

@hjjq Could you double check whether we need this? Thanks!

yaoyaoding (approved):
Thanks, @hjjq. LGTM. Only leave some minor issues. 

Good job!

PR #88, opened 2023-02-12
[Bugfix] 'double'->'float64' in onnx dtype conversion

### LINKED ISSUE #87: [Bug] Data type casting from onnx
When importing an onnx model, hidet needs to map onnx data types into hidet dtypes.

However, in

https://github.com/hidet-org/hidet/blob/0cd78c7828232062992d9d5429fd5b0179bf2daa/python/hidet/graph/frontend/onnx/onnx.py#L485

hidet does not have dtype ``double``. I guess it should be ``float64``?

soodoshll:
fix #87


yaoyaoding:
Thanks @soodoshll.

PR #125, opened 2023-03-02
[Operator] Move task building into task method to allow for override

xinli-git:
I am working on a new functionality of hidet that allows external registration of tasks that is not built using `implement_cuda` nor `auto_schedule`. This allows for the following possible use cases

- certain tasks may require implementations that cannot be expressed (or too difficult) using hidet script or hidet compute expression. One example [Non-max suppresion](https://github.com/onnx/onnx/blob/main/docs/Operators.md#NonMaxSuppression)
- other libraries may have sophisticated optimizations that would allow fast and convenient integration to Hidet as tasks

To achieve this, the `build` routine for those tasks requires customization. Previously hidet assumes the same build routine and leave that as an external function. This change moves the build into a class method to allow for overrides.


xinli-centml:
close this PR as Yaoyao is already working on it

PR #126, opened 2023-03-03
[Dynamo] Sigmoid and Hardsigmoid operator and module support

AndreSlavescu:


tests/frontends/torch/test_torch_hardsigmoid.py
```diff
@@ -0,0 +1,35 @@
...
+def test_flatten(shape):
+    hardsigmoid = HidetHardsigmoid(torch.nn.Hardsigmoid())
+    check_torch_unary(
+        shape,
+        lambda x: torch.nn.functional.hardsigmoid(x),
+        lambda x: hardsigmoid(x),
+        dtype='float32',
+        atol=1e-5,
+        rtol=1e-5,
+    )
```
yaoyaoding, line 31: It is better to test new hidet operator (under hidet.graph.ops) in ``tests/graph/operators`` in this way. 

To test the translation of pytorch model to hidet graph, it is better to
1. define a torch model (in this case, a model with a single operator that uses torch.nn.HardSigmoid),
2. run the model with hidet dynamoc backend,
3. compare the result with pytorch eager mode.

yaoyaoding (changes requested):
Hi @AndreSlavescu, thanks for the PR.

You have added the tests for the new hardsigmoid itself, but did not test the translation from pytorch operator to hidet operator. Could you add a test for this?

PR #135, opened 2023-03-21
[Graph] Cast optimizations

xinli-git:
Add 3 graph patterns to eliminate redundant casts 

- y1 = cast(x), y2 = cast(x) => y1 = y2 = z = cast(x)
- same as above but for 3 outputs
- cast(cast(x)) where the outer casts has the same dtype as x => x


yaoyaoding:
Thanks, @xinli-git!

PR #146, opened 2023-03-24
[Operator] Resolve matmul to batch_matmul when lower than sm80

hjjq:
@yaoyaoding Let me know if you think Hidet should have a different behavior.


yaoyaoding:
Thanks @hjjq. That's also what I think for now.

PR #153, opened 2023-04-03
[Dynamo] Fix dynamo tests and dump graph IR

xinli-git:
It seems that torch.compile on a single torch.nn.* operator will not trigger compilation. 

Use a temporary hack to wrap module by `torch.nn.Sequential` until a proper fix is introduced. 

Now that this test is enabled, operators added previously for testing actually needed hidet support. Also fix some operator params that was misconfigured. 

There was also a small bug where save_graph_ir was not called in PassContext for Dyamo, adding it here.


yaoyaoding:
Thanks @xinli-git !

PR #173, opened 2023-04-13
[Dynamo] Operator Mappings

AndreSlavescu:
- mapping for operator.neg
- mapping for torch.gather


yaoyaoding:
The `torch.gather` (aka `hidet.ops.gather`) operator has been added in #174, close this for now.

PR #176, opened 2023-04-15
Refactor codegen to separate GPU/CPU code generation

LDY1998:
This is an attempt to refactor codegen.py to separate code generation for GPU and CPU, it also adds emulation for fp16 and bfloat16 as cpu has no native support for them. I didn't change the workflow of compilation yet since we only want to run CPU codegen of hidet in an environment without CUDA support, this means that all the dependencies of `cudart` should not run when GPU is not available.

Tasks:
- [x] Refactor codegen
- [x] Add fp16 and bf16 emulation

Tasks to be done:
- [] Separate dependencies of `cudart`
- [] Change to compilation workflow to run CPUCodegen when GPU is not available


python/hidet/ir/primitives/cpu/math/bfloat16.py
```diff
@@ -0,0 +1,98 @@
...
+                codegen_name=codegen_name,
+                func_or_type=FuncType(param_types=['bfloat16'] * num_args, ret_type='bfloat16'),
+            )
+
+    def call(self, name: str, *args) -> Expr:
+        entry = primitive_func_pool.lookup_by_name(name)
+        return Call(entry.var, args)
+
+    def sin(self, a: Expr) -> Expr:
+        return self.call('cpu_bf16_sin', a)
```
yaoyaoding, line 51: Might be better to make the name conversion consistent with the float16?

```suggestion
        return self.call('sin', a)
```
and update the `self.call(...)`

python/hidet/backend/codegen.py
```diff
@@ -663,8 +534,275 @@ def visit_AnyExpr(self, e: AnyExpr):
...
+        if module.task is not None:
+            doc += '/*' + NewLine()
+            doc += str(module.task) + NewLine()
+            doc += '*/' + NewLine()
+
+        doc += Text('extern "C" {') + NewLine()
+
+        # add namespace to activate data type and function
+        doc += Text('using float16::Half;') + NewLine()
+        doc += Text('using bfloat16::BFloat16;') + NewLine()
```
yaoyaoding, line 762: Besides `using float16::Half;`, do we need also add a typedef like the following?
```
typedef Half half;
typedef BFloat16 bfloat16_t;
```

include/hidet/cpu/bfloat16.h
```diff
@@ -0,0 +1,686 @@
...
+        {
+            ufrom.i--;
+        }
+        else
+        {
+            ufrom.i++;
+        }
+
+        return ufrom.f;
+    }
```
yaoyaoding, line 481: Anywhere this function is used?

include/hidet/cpu/bfloat16.h
```diff
@@ -0,0 +1,686 @@
...
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
```
yaoyaoding, line 78: Ideally, we hope not to introduce external source code with restrictive license. We might want to replace these float16 and bfloat16 implementation with ours. But for now, it is okay use to use them.

yaoyaoding (commented):
Hi @LDY1998, the PR looks good in general. I left some minor comments. Because it does not influence the existing cuda implementation, after you fixing these comments, we can merge it to the main branch.

yaoyaoding:
You can also create an issue to track the support for CPU backend.

python/hidet/backend/codegen.py
LDY1998: You are right, I will add the typedef

include/hidet/cpu/bfloat16.h
LDY1998: It's not used in hidet for now, I will remove it from the header.

yaoyaoding:
Thanks @LDY1998!

PR #189, opened 2023-04-26
Refactor compilation workflow to support CPU without CUDA

LDY1998:


yaoyaoding:
Thanks @LDY1998! I also updated part of the implementation, could you have a look?

LDY1998 (commented):
Thanks @yaoyaoding for the update. I only left a very minor comment. I think overall it's good to go.

python/hidet/backend/build.py
```diff
@@ -39,20 +39,163 @@ def __str__(self):
...
+                # if the compilation failed, raise an exception
+                if result.returncode:
+                    message = "Command: {}\n".format(command)
+                    if result.stdout:
+                        message += result.stdout.decode().strip() + '\n'
+                    if result.stderr:
+                        message += result.stderr.decode().strip()
+                    raise CompilationFailed(src_path, message)
+
+                # write the compilation log to "nvcc_log.txt"
```
LDY1998, line 74: Do we want to leave the comment like that

yaoyaoding:
Thanks @LDY1998.

Issue #217, opened 2023-05-08
[Bug] Hidet require more GPU memory than native torch compile


eric8607242:
Hi, 
thanks for the great work again!

**Describe the bug**
Currently, I want to compile the LLama-7B on a Nvidia 4090 GPU using Hidet for faster inference speed.
However, I have encountered an issue where Hidet requires more GPU memory compared to the native torch.compile method.

For example, given a `input_ids` with shape `[2, 128]`. The native `torch.compile` can compile the model successfully and infer the output correctly, but, when using Hidet, an OOM exception is raised during the first inference step.

I am wondering why Hidet requires more GPU memory.
Do I need some additional configurations to save the GPU memory?

**To Reproduce**
The following script is the example code to compile the LLaMA.
```python
import torch
from transformers import LlamaModel, LlamaConfig
import hidet

print("Initialize the model")
configuration = LlamaConfig()
model = LlamaModel(configuration).half().eval().cuda()

BATCH_SIZE = 2
SEQ_LEN = 128
input_ids = torch.zeros(2, 128, dtype=torch.long).cuda()

with torch.no_grad():
    hidet.torch.dynamo_config.use_tensor_core(True)
    hidet.torch.dynamo_config.search_space(2) 
    hidet.torch.dynamo_config.use_fp16(True)
    hidet.torch.dynamo_config.use_fp16_reduction(True)

    print("Start to compile")
    # Compile the model using Hidet
    model_opt = torch.compile(model, backend='hidet')
    #model_opt = torch.compile(model, mode="reduce-overhead")
    print("Start to inference")
    model_opt(
        input_ids=input_ids,
        output_hidden_states=True
    )
```
However, the current Hidet has some unsupported operations for [huggingface LLaMA](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py).
I fork the repo and add some operations based on the latest version in this [branch ](https://github.com/eric8607242/hidet/tree/llama_complementary), which can compile the model successfully.

**Enviroment**
- Python Package Requirements
```
transformers == 4.28.1
torch == 2.0.0
sentencepiece == 0.1.99
```
- System Enviroments
    - OS: Ubuntu 22.04.1 LTS
    - GPU: RTX 4090
    - CUDA Version: 11.8
    - Driver Version: 520.61.05

yaoyaoding:
Hi @eric8607242,

Glad to see that you have tried this model and added some operators by yourself! 

[...]

PR #226, opened 2023-05-13
[Example] Basic stable diffusion demo

Aalanli:
A basic stable diffusion ui with gradio, however, eager and compiled modes have the same timing on my machine.

### DIFF
--- a/examples/stable-diffusion/main.py
+++ b/examples/stable-diffusion/main.py
@@ -1,0 +1,122 @@
+import gradio as gr
+import torch
+from torch import autocast
+from diffusers import DiffusionPipeline
+from diffusers import DPMSolverMultistepScheduler
+
+import time
+import hidet
+
+
+model_id = "CompVis/stable-diffusion-v1-4"
+device = "cuda"
+
+dpm = DPMSolverMultistepScheduler.from_pretrained(model_id, subfolder="scheduler")
+pipe = DiffusionPipeline.from_pretrained(model_id, torch_dtype=torch.half, scheduler=dpm).to("cuda")
+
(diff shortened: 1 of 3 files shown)


Aalanli:
With the default settings.

yaoyaoding:
Hi @Aalanli, thanks for the PR!

Let's wait for the support for the scaled_dot_product_attention. 

I have run this demo locally, and we will have out of memory issue (on my 3090) when we tried to tune in space 1 or 2. The reason might be the same as https://github.com/hidet-org/hidet/issues/217. We can merge this PR first (after we have scaled_dot_product), and then consider construct the model like https://github.com/hidet-org/hidet/blob/main/python/hidet/testing/models/resnet.py (use the weights from diffusers).

examples/stable-diffusion/main.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        time_text += f"eager time: {eager_time}"
+    if compile_time == 0:
+        time_text += " | compile time: ..."
+    else:
+        time_text += f" | compile time: {compile_time}"
+    
+    return images, time_text
+
+def adjust_compile_opts(search_space=0, fp16=True, tensor_core=True, fp_16_red=False, cuda_graph=True, attn=False):
```
xinli-git, line 56: attn is not used here? Should we give it a shot and see if the pictures are OK?

Aalanli: Yes, I fixed a minor issue, but the pictures should be alright, at least with the default settings.

xinli-centml: Thanks!

PR #227, opened 2023-05-15
fix batch_matmul for invalid mma config for sm  < 80

xinli-git:
Some MMA configs are not valid in sm<80. Fix this by filtering out mma_configs before creating the tuning space.

This should address #225 

Also, I changed the warning for contiguous tensor to warn_once to reduce the amount of context printed to the consol.


xinli-git:
There is another minor annoyance with tuning this model. By default, hidet parallel tuning/build uses # of cores as parallel workers. However, when the fused kernel is large, the compilation processes would use all the memory in the system and cause system hang.

Setting `hidet.option.parallel_build(enabled=True)` seems to have no effect. 

I will make a fix in another PR

yaoyaoding:
Thanks @xinli-git, it looks good to me.

xinli-git:
Thanks @yaoyaoding, I will work on this next :)

PR #243, opened 2023-05-23
[Model] Llama model implementation in hidet

Aalanli:
Preliminary llama model implementation in hidet. 
Changed some imports to make the structure more like torch for convenience.

Verified model loading on a small synthetic model. However, I cannot verify the actual model loading due to memory constraints.

Inference script not included, working on auto translator to move the huggingface infrastructure over from torch.


python/hidet/graph/tensor.py
```diff
@@ -1468,3 +1510,26 @@ def asarray(obj, /, *, dtype=None, device=None) -> Tensor:
...
+
+
+def matmul(a: Tensor, b: Tensor) -> Tensor:
+    """
+    Performs matrix multiplication between a and b
+
+    """
+    from .ops import matmul as _matmul
+
+    return _matmul(a, b)
```
yaoyaoding, line 1535: Hi @Aalanli , is it necessary to place them here? I prefer we leave the definitions of these two functions to `hidet.ops.*`, and directly use `hidet.ops.concat` and hidet.ops.matmul`.

python/hidet/testing/models/llama.py
```diff
@@ -0,0 +1,668 @@
...
+        >>> tokenizer = AutoTokenizer.from_pretrained(PATH_TO_CONVERTED_TOKENIZER)
+
+        >>> prompt = "Hey, are you consciours? Can you talk to me?"
+        >>> inputs = tokenizer(prompt, return_tensors="pt")
+
+        >>> # Generate
+        >>> generate_ids = model.generate(inputs.input_ids, max_length=30)
+        >>> tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False)[0]
+        "Hey, are you consciours? Can you talk to me?\nI'm not consciours, but I can talk to you."
+        ```"""
```
yaoyaoding, line 518: We do not have these functionality anymore after ported to hidet, let's remove them :)

python/hidet/testing/models/llama.py
```diff
@@ -0,0 +1,668 @@
...
+    # Get a list of parameters in advance, then delete the model to save memory
+    param_list = [param for _, param in hf_model.named_parameters()]
+    del hf_model
+
+    for i, param in enumerate(param_list):
+        param_list[i] = hidet.from_torch(param.detach()).to(config.dtype)
+
+    # construct the new model without duplicating the weights
+    # since the hidet model is isomorphic to the original torch model
+    # this only works if we add .register_buffer(...)
```
yaoyaoding, line 650: I saw you added `register_buffer`, do we still need this?

yaoyaoding (commented):
Hi @Aalanli,

I hvae taken a rough look and it looks good. One minor concern is that whether we need to add the `matmul` and `cat` under the `hidet.graph.tensor` namespace. I still prefer we directly use the `hidet.ops.matmul` and `hidet.ops.concat` instead.

yaoyaoding:
By the way, I have refactored the structure of `hidet.graph.nn`. Could you resolve the caused conflicts? Thanks!

python/hidet/graph/tensor.py
Aalanli: Ok, will change.

python/hidet/testing/models/llama.py
Aalanli: Right.

python/hidet/testing/models/llama.py
Aalanli: If we leave it out, the registered buffers will be under the same namespace as the regular parameters in the hidet model, but not the torch one, so we would need to implement model.named_parameters() like torch, to reuse the parameters, which would be more complex. Is there another way otherwise?

yaoyaoding: I have implemented the `named_parameters(...)` in the latest commit: https://github.com/hidet-org/hidet/blob/main/python/hidet/graph/nn/module.py#L96-L103

Does it work for this usage case?

python/hidet/graph/tensor.py
yaoyaoding: Thanks, just want to keep the `hidet.graph.tensor` namespace clean.

python/hidet/testing/models/llama.py
Aalanli: For example, 
https://github.com/hidet-org/hidet/blob/6d4bd3d531ba5a839269e3f648baec8179d04f59/python/hidet/testing/models/gpt2.py#L207-L242

constructs the model twice in memory.

yaoyaoding: Good catch!

We can first load the hf model in cpu.

Aalanli:
Just waiting for the graph fusion failure to be addressed, currently generating prompts with no graph optimization works as expected.

Aalanli:
Waiting for PR#268

yaoyaoding:
Thanks @Aalanli !

PR #257, opened 2023-05-29
[Operator] Add new operator "normalize" that makes a group of layers (layer norm, group norm and instance norm) faster using hidet script

xinli-git:
Refactor the normalize op to use a dedicated task and hidet script implementation

3 main optimizations
* [welford reduce](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance): this is a one-pass algorithm that only reads from global memory once and writes it back once. Compared to the previous approach, this will reduce global memory by at least 2x.
* fused mean var calculation and norm into a single kernel -> reduce redundant global memory access
* 2 stage reduce (1024 -> 32 -> 1) for larger reduction sizes

Before
```
Compiling cuda task subtract(x=float32(1, 128, 768), y=float32(1, 128, 1), z=float32(1, 128, 768))...
Compiling cuda task reduce_avg(x=float32(1, 128, 768), y=float32(1, 128, 1), dims=[2], keep_dim=True, reduce_type='avg', accumulate_dtype='float32')...
Compiling cuda task fused(x=float32(1, 128, 768), y=float32(1, 128, 1), fused_ops='square reduce_avg adds rsqrt', anchor='reduce_avg')...
Compiling cuda task fused(y=float32(768,), x=float32(1, 128, 768), y=float32(1, 128, 1), z=float32(1, 128, 768), fused_ops='mul mul', anchor='mul')...
Compiling cuda task fused(y=float32(768,), y=float32(768,), x=float32(1, 128, 768), y=float32(1, 128, 1), z=float32(1, 128, 768), fused_ops='mul mul add', anchor='add')...
```

now:
```
Compiling cuda task normalize_float32(x=float32(1, 128, 768), y=float32(1, 128, 768), dims=[2], accumulate_dtype='float32', epsilon=1e-12f)...

```

I have a suspicion that this can somehow also work with epilogue fusion enabled, but probably need to discuss further with Yaoyao. The use case is that usually there is an affine transformation after the normalization.  

I tried the benchmark on bert-base-uncased (resnet is unaffected), there was not a huge improvement in performance, I think this model is not heavily limited by layernorm. I'll look at some micro benchmarking soon. 

```bash
hidet bench --cache-dir ./test_cache --space 2 --dtype float32 --report bert-seq128-f32.txt --tensor-core nlp --seq-length 128 --models bert-base-uncased
...
| model                   | inputs             |   eager |   reduce-overhead |   max-autotune |   hidet(2) |

|-------------------------|--------------------|---------|-------------------|----------------|------------|

| model/bert-base-uncased | f32, bs=1, seq=128 |   2.863 |             2.715 |          2.815 |      2.023 |

```
and 
```bash
hidet bench --cache-dir ./test_cache --space 2 --dtype float16 --report bert-seq128-f16.txt --tensor-core nlp --seq-length 128 --models bert-base-uncased
| model                   | inputs             |   eager |   reduce-overhead |   max-autotune |   hidet(2) |

|-------------------------|--------------------|---------|-------------------|----------------|------------|

| model/bert-base-uncased | f16, bs=1, seq=128 |   1.710 |             1.701 |          1.097 |      1.521 |
```

As compared to main
```
FP32
| model                   | inputs             |   eager |   reduce-overhead |   max-autotune |   hidet(2) |

|-------------------------|--------------------|---------|-------------------|----------------|------------|

| model/bert-base-uncased | f32, bs=1, seq=128 |   3.043 |             2.858 |          2.779 |      2.076 |

| model                   | inputs             |   eager |   reduce-overhead |   max-autotune |   hidet(2) |

FP16

|-------------------------|--------------------|---------|-------------------|----------------|------------|

| model/bert-base-uncased | f16, bs=1, seq=128 |   1.704 |             1.671 |          1.098 |      1.573 |
```


tests/frontends/torch/test_torch_norm.py
```diff
@@ -18,14 +18,21 @@
...
 @pytest.mark.parametrize('shape', [[1, 4, 32, 32]])
 @pytest.mark.parametrize('num_groups', [1, 2, 4])
 @pytest.mark.parametrize('dtype', [torch.float32])
 def test_group_norm(shape, num_groups, dtype):
-    check_module(torch.nn.GroupNorm(num_groups=num_groups, num_channels=shape[1]), [torch.randn(shape, dtype=dtype)])
+    check_module(
+        torch.nn.GroupNorm(num_groups=num_groups, num_channels=shape[1]),
+        [torch.randn(shape, dtype=dtype)],
+        atol=1e-2,
+        rtol=1e-2,
```
yaoyaoding, line 34: Does current implementation pass the atol=rtol=1e-4 precision?

python/hidet/graph/ops/definitions/normalize/resolve.py
```diff
@@ -0,0 +1,62 @@
...
+    Resolve a generic reduce operator according to the following rules, in decreasing priority:
+    1) resolve_f16: If the input data type is float16, and the size of the last input dimension is an even number,
+        return the output of the f16 optimized reduce schedule. (Support of odd number will be added in the future)
+    2) resolve_generic: Default case, return the output of the regular f32 reduce schedule.
+    """
+
+    def resolve_f16(self, op: Operator) -> Optional[List[Tensor]]:
+        dims = op.attrs['dims']
+        x: Tensor = op.inputs[0]
+        if not is_contiguous_norm(dims, len(x.shape)):
```
yaoyaoding, line 37: Let's discuss offline whether it is possible to implement the version that support incontiguous dimentions.

python/hidet/graph/ops/definitions/normalize/norm_f16.py
```diff
@@ -0,0 +1,272 @@
...
+                    + delta
+                    * delta
+                    * cast(count_a[0], accumulate_dtype)
+                    * cast(count_b[0], accumulate_dtype)
+                    / cast(count, accumulate_dtype)
+                )
+                count_a[0] = count
+
+            @hidet.script
+            def norm_kernel(x: f16[x.const_shape], y: f16[y.const_shape]):
```
yaoyaoding, line 92: We might want to read float16x2 at a time in the future instead of single float16, the former one would have better efficiency.

python/hidet/graph/ops/definitions/normalize/norm_f16.py
```diff
@@ -0,0 +1,272 @@
...
+from hidet.lang import data_type, TensorType, i32, f16, attrs, tensor
+from hidet.lang.cuda import blockIdx, threadIdx, register_tensor, syncthreads
+from hidet.graph.ops.definitions.utils import Operator, Tensor, normalize_dim
+from hidet.graph.ops.definitions.utils import input_like
+from hidet.utils import prod
+from hidet.ir import primitives as prim
+from .norm import NormalizeTask
+
+
+class NormalizeF16Task(NormalizeTask):
```
yaoyaoding, line 25: I am not sure which one is better: implement a fp16 version seperately vs. implement the fp16 & fp32 version in the same template.

yaoyaoding (approved):
Thanks @xinli-git ! Great job!

I left some comments, and we can discuss them offline. This PR itself is ready be to merged after passing the CI.

tests/frontends/torch/test_torch_norm.py
xinli-git: removed

python/hidet/graph/ops/definitions/normalize/norm_f16.py
xinli-git: with 2xf16 it is probably better to keep them seperate. I will add it.

Otherwise most of the logic is the same and we can basically template on <input dtype, accumulate dtype>

python/hidet/graph/ops/definitions/normalize/norm_f16.py
xinli-git: you are right, this approach currently only loads 64 bytes coalesced.  I will add it

xinli-git:
Hi Yaoyao, I will merge this for now and leave the performance enhancements  in a separate PR

PR #258, opened 2023-05-29
[Fixbug] Fix test warnings and the incompatibility of two recent PRs

yaoyaoding:
Fix test warnings and allow the norm support dynamic shape.


yaoyaoding:
Thanks @xinli-git for the help!

PR #268, opened 2023-06-02
[Dynamic Shape] Adding dynamic shape support for reduce

Aalanli:
Dynamic shape support for reduce is required by llama.

However, this introduces Expr into __launch_bounds__, so I add a check for that case.

Future work:
- add more tests for dynamic shape
- add more dynamic shape support for more operators
- add assert feature in hidet script, as to assert things during kernel launch in compiled kernel, due to dynamic shape (see reduce_fp16.py, which checks that the last dimension is divisible by 2)


python/hidet/backend/codegen.py
```diff
@@ -715,11 +715,13 @@ def visit_Function(self, func: Function) -> Doc:
...
-                doc += f' __launch_bounds__({block_dim}, {min_blocks})'
-            else:
-                doc += f' __launch_bounds__({block_dim})'
+            if isinstance(block_dim, (Constant, int)):
+                if 'cuda.min_blocks' in func.attrs:
+                    min_blocks = func.attrs['cuda.min_blocks']
+                    if isinstance(min_blocks, (Constant, int)):
+                        doc += f' __launch_bounds__({block_dim}, {min_blocks})'
+                else:
+                    doc += f' __launch_bounds__({block_dim})'
```
yaoyaoding, line 724: When block_dim is constant, 'cuda.min_blocks' in func.attrs but 'cuda_min_blocks' is not constant, we should also set the launch bound for block_dim.

python/hidet/graph/ops/definitions/reduce/reduce_f16.py
```diff
@@ -293,6 +293,12 @@ def __init__(self, x: Tensor, dims: Optional[Sequence[int]], keepdims: bool = Fa
 def reduce_f16(
     x: Tensor, dims: Union[int, Sequence[int]], keepdims: bool, reduce_type: Union[ReduceType, str]
 ) -> Tensor:
+    from hidet.ir import Expr
+
+    if True in [isinstance(i, Expr) for i in x.shape]:
+        # TODO: either remove x.shape[-1] % 2 != 0 condition
+        # or for more longterm solution, add conditional statements to kernel launch
```
yaoyaoding, line 300: The best solution is to add a schedule template that support x.shape[-1] % 2 == 1.

Before that, we can add the constraint in our resolve rule for reduce operator, to fall back to default implementation in case x.shape[-1] is symbolic or odd number.

python/hidet/graph/ops/definitions/reduce/reduce_f16.py
```diff
@@ -293,6 +293,12 @@ def __init__(self, x: Tensor, dims: Optional[Sequence[int]], keepdims: bool = Fa
 def reduce_f16(
     x: Tensor, dims: Union[int, Sequence[int]], keepdims: bool, reduce_type: Union[ReduceType, str]
 ) -> Tensor:
+    from hidet.ir import Expr
+
+    if True in [isinstance(i, Expr) for i in x.shape]:
```
yaoyaoding, line 298: Personally, I prefer 
```suggestion
    if any(isinstance(d, Expr) for d in x.shape):
```

yaoyaoding (commented):
Thanks @Aalanli! 

Overall LGTM, left a few minor comments.

python/hidet/backend/codegen.py
yaoyaoding:
```suggestion
            if isinstance(block_dim, (Constant, int)):
                if 'cuda.min_blocks' in func.attrs:
                    min_blocks = func.attrs['cuda.min_blocks']
                    if isinstance(min_blocks, (Constant, int)):
                        doc += f' __launch_bounds__({block_dim}, {min_blocks})'
                    else:
                        doc += f' __launch_bounds__({block_dim})'
                else:
                    doc += f' __launch_bounds__({block_dim})'
```

PR #273, opened 2023-06-05
[Tools] Model translator

Aalanli:
See README and examples


python/hidet/utils/model_translator/utils.py
```diff
@@ -0,0 +1,31 @@
+# %%
```
yaoyaoding, line 1: We also have the color support at https://github.com/hidet-org/hidet/blob/a1bd18cac4b959d63a8bc7c371e174d3cabe81ca/python/hidet/utils/py.py#L61-L121

We can consider reusing it.

python/hidet/utils/model_translator/tests/classes.py
```diff
@@ -0,0 +1,91 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
yaoyaoding, line 1: Better to move the tests `classes.py` and `functions.py` to `tests/utils/model_translator/` folder.

yaoyaoding (approved):
Hi @Aalanli,

Thanks for the great tool! I left some minor comments.

PR #274, opened 2023-06-07
[Operator] Add dynamic shape support and tests for Operators.

Aalanli:
Altered operators and added the associated test for dynamic shape support.

Changed:
- conv1d / conv1d_transposed
- conv2d / conv2d_transposed
- conv3d / conv3d_transposed
- batch_matmul
- matmul_fp16
- softmax

However, there are certain shape checks that are not possible for dynamic dimensions during compile time, I have skipped over those for now; the shape checks will not happen if its dynamic, but I think its a definite must for the future.

Some difficulties arise comes when resolve rules rely on shapes, short-term fix would be to resolve to default implementation if shape is dynamic, long term would probably be to add conditionals to graph IR, however, this may not play very nicely with fusion...

Similarly, long term fix for operators and task mappings would probably be to add some sort of 'assert' node to the IR, so that assertions gets added to the launch function in codegen.


yaoyaoding:
Thanks @Aalanli !

PR #276, opened 2023-06-09
[Dynamic Shape] More correctness guards

Aalanli:
Add support for dynamic shape assertions in the C++ runtime.
Further, add shape and dtype check for python runtime. With this being enabled by default.


Aalanli:
Moreover, there is a subtle bug where the symbol registry gets rewritten multiple times if there exists inputs which promises the same shape, but runtime inputs have different shapes.
Eg. hidet.symbol(['a', 'a']), but during runtime: hidet.randn([1, 2])

So added a check for that as well.

Aalanli:
I have no idea why the gpt2 test passed before, there was a similar error with the llama implementation, where the compiled model reinterpreted an int64 tensor as an int32 tensor, producing random outputs.

python/hidet/ir/tools/printer.py
```diff
@@ -474,6 +474,15 @@ def visit_Task(self, e: Task):
             Text('computations: ') + self.print_tensor_nodes(e.outputs).indent(),
             Text('attributes: {') + self({k: str(v) for k, v in e.attrs.items()}) + '}',
         ]
+        if len(e.assertions) > 0:  # between computations and attributes
+            lines.insert(
+                -1,
```
yaoyaoding, line 479:
```suggestion
            lines.append(
```

python/hidet/option.py
```diff
@@ -563,6 +570,30 @@ def debug_show_var_id(enable: bool = True):
...
+    Whether to check shapes and dtypes of all input arguments to compiled Graphs or Tasks.
+
+    Parameters
+    ----------
+    enable: bool
+        Whether to check shapes and dtypes of all input arguments to compiled Graphs or Tasks.
+    """
+    OptionContext.current().set_option('runtime_check', enable)
+
+
```
yaoyaoding, line 596: To be consistent with other option API, better to design these two as
```python
def runtime_check(enable: bool = True):
    pass
def get_runtime_check() -> bool:
    pass
```

python/hidet/runtime/compiled_graph.py
```diff
@@ -277,6 +277,33 @@ def run_async(self, inputs):
         ret: List[hidet.Tensor]
             The output tensors.
         """
+        if hidet.option.runtime_check():
+            symbol_map = {}
+            for i, (traced, new) in enumerate(zip(self.meta.inputs, inputs)):
```
yaoyaoding, line 282:
```suggestion
            for i, (expected, actual) in enumerate(zip(self.meta.inputs, inputs)):
```

python/hidet/runtime/compiled_graph.py
```diff
@@ -277,6 +277,33 @@ def run_async(self, inputs):
...
+                    raise RuntimeError(
+                        f"dtype mismatch at arg {i} between original: {traced.dtype} and new: {new.dtype}"
+                    )
+                traced_shape = traced.shape
+                concrete_shape = new.shape
+                if len(traced_shape) != len(concrete_shape):
+                    raise RuntimeError(
+                        f"Rank of input {i} not equal to original. ({len(concrete_shape)} vs. {len(traced_shape)})"
+                    )
+                for j, (orig_shape, new_shape) in enumerate(zip(traced_shape, concrete_shape)):
```
yaoyaoding, line 293:
```suggestion
                for j, (expected_size, actual_size) in enumerate(zip(traced_shape, concrete_shape)):
```
The shape usually means all dimension sizes

python/hidet/runtime/compiled_graph.py
```diff
@@ -277,6 +277,33 @@ def run_async(self, inputs):
...
+                                           original: {orig_shape} vs. new: {new_shape}'
+                        )
+                    elif orig_shape not in symbol_map:
+                        symbol_map[orig_shape] = new_shape
+                    elif symbol_map[orig_shape] != new_shape:
+                        raise RuntimeError(
+                            f"There exists multiple instances of the same symbol {orig_shape} with\
+                            different values in inputs (ex: {symbol_map[orig_shape]} and {new_shape})"
+                        )
+
```
yaoyaoding, line 306: Implement the checking as a new method (say `_check_inputs`).

python/hidet/runtime/compiled_task.py
```diff
@@ -147,6 +147,35 @@ def pick_best_candidate(self, inputs, outputs) -> int:
...
+                            f'shape mismatch at dimension {j}, original: \
+                                           {orig_shape} vs. new: {new_shape}'
+                        )
+                    elif orig_shape not in symbol_map:
+                        symbol_map[orig_shape] = new_shape
+                    elif symbol_map[orig_shape] != new_shape:
+                        raise RuntimeError(
+                            f"There exists multiple instances of the same symbol {orig_shape}\
+                            with different values in inputs (ex: {symbol_map[orig_shape]} and {new_shape})"
+                        )
```
yaoyaoding, line 177: Same logic as the CompiledGraph. Consider implement as a utility function and put it at the same module of `TensorSignature`.

yaoyaoding (commented):
Thanks @Aalanli.

yaoyaoding:
Thanks @Aalanli !

PR #280, opened 2023-06-16
[Distributed] add nccl primitives

soodoshll:
What we have now:

* automatically finding headers and libraries provided by the pypi package `nvidia-nccl-cu11` or `nvidia-nccl-cu12`
* expose nccl's initialization apis to the python runtime
* a python launch script to start single-machine-multi-gpu tasks (`./examples/distributed/test.py`)
* modified cudacontext to include a pool of nccl communicators
* primitive for all_reduce


python/hidet/ffi/ffi.py
```diff
@@ -40,6 +42,18 @@ def load_library():
...
+    library_dirs = get_nccl_library_search_dirs()
+    for library_dir in library_dirs:
+        lib_nccl_paths = glob.glob(os.path.join(library_dir, 'libnccl.so*'))
+        if len(lib_nccl_paths) == 0:
+            continue
+        _LIB_NCCL = ctypes.cdll.LoadLibrary(lib_nccl_paths[0])
+        library_paths['nccl'] = lib_nccl_paths[0]
+        break
+    if _LIB_NCCL is None:
+        raise OSError('Can not find nccl library in the following directory: \n' + '\n'.join(library_dirs))
```
yaoyaoding, line 56: Let's put this part in "hidet/cuda/nccl/ffi.py", and leave "hidet/ffi/ffi.py" to only contain the hidet runtime library.

In the future, when we want to add other library (e.g., cudnn library), we can put to "hidet/cuda/cudnn" and also its ffi.py.

python/hidet/ffi/runtime_api.py
```diff
@@ -9,11 +9,43 @@
...
+        runtime_api.add_nccl_comm(self)
+    
+    def __del__(self):
+        """
+        Should we manage the lifetime of communicator object in Python or C++?
+        """
+        nccl_runtime_api.comm_destroy(self)
+
+    def split(self):
+        raise NotImplementedError()
```
yaoyaoding, line 48: Consider moving this to "hidet/cuda/nccl.py".

python/hidet/ffi/runtime_api.py
```diff
@@ -61,5 +95,63 @@ def set_symbol_value(name: str, value: int) -> None:
...
+        @staticmethod
+        def comm_destroy(comm:NcclCommunicator) -> None:
+            ret = NCCLRuntimeAPI._comm_destroy(comm._handle)
+            assert ret == 0
+
+    nccl_runtime_api = NCCLRuntimeAPI()
+    _comms: List[NcclCommunicator] = []
+
+    def get_nccl_comm(comm_id: int):
+        return _comms[comm_id]
```
yaoyaoding, line 157: Also move to `hidet.cuda.nccl`

python/hidet/libinfo.py
```diff
@@ -66,3 +66,13 @@ def get_library_search_dirs() -> List[str]:
...
+
+def _get_nccl_dirs():
+    import site
+    return [os.path.join(root, 'nvidia', 'nccl') for root in site.getsitepackages()]
+
+def get_nccl_include_dirs():
+    return [os.path.join(root, 'include') for root in _get_nccl_dirs()]
+
+def get_nccl_library_search_dirs():
+    return [os.path.join(root, 'lib') for root in _get_nccl_dirs()]
```
yaoyaoding, line 78: Move to `hidet.cuda.nccl`

yaoyaoding (commented):
Hi @soodoshll , thanks!

Leave some comments on how to organize the nccl-related source code. I prefer to putting them under `hidet.cuda.nccl` submodule. This would make our code structure more clean when we add other vendor libraries like cublas and cudnn.

python/hidet/ffi/runtime_api.py
yaoyaoding: Good questions!

> How should we manage the lifetime of NcclCommunicator objects in Python and ncclComm_t objects in C++?

Because different flow graphs may have different set of communicators, thus, we should create the communicators for each flow graph. Let's keep the communicators in CompiledGraph. Something like

```python
class CompiledGraph:
def __init__(...):
    ...
    self.nccl_comms = ... # create the communicators, 
                          # let's store it as hidet.ffi.utils.Array(void*, num of comm)
                          # and create a runtime api called 
                          # "void set_nccl_comms(int num_comm, void** comm_array)"
def run_async(...):
    ...
    runtime_api.set_nccl_comms(len(self.nccl_comms), self.nccl_comms)
    ...

def __del__(...):
    # destroy communicators
```

For `FlowGraph.forward(...)`, we can raise an error for now if we find it is a distributed flow graph. 

> Should we import 'nccl.py' in 'nccl/ffi.py' or the opposite? Because NcclCommunicator might depend on some runtime api, therefore we need 'nccl/ffi.py' in 'nccl.py' which might cause circular import.

Consider: let the nccl runtime api return integer (the pointer to communicator), and the `hidet/cuda/nccl/nccl.py` imports `hidet/cuda/nccl/ffi.py`. The nccl ffi should only be used by `nccl.py` and we expose the api to the users in `nccl.py`.

python/hidet/ir/primitives/cuda/nccl.py
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+from hidet.ir.expr import Expr
+from hidet.ir.stmt import BlackBoxStmt
+
+# TODO: we should not put nccl-related types here since hidet.cuda.nccl depends on
+#       the existence of nccl library?
+from hidet.cuda.nccl import NcclDataType, NcclRedOp
+
+
+def all_reduce(comm_id: int, sendbuff: Expr, recvbuff: Expr, count: Expr, dtype: NcclDataType, op: NcclRedOp):
```
yaoyaoding, line 20:
```suggestion
def all_reduce(comm_id: int, sendbuff: Expr, recvbuff: Expr, count: Expr, dtype: DataType, op: NcclRedOp):
```
Let's pass the `hidet.ir.type.DataType` to this primitive and convert it into NcclDataType inside the primitive function.

src/hidet/runtime/cuda_context.cpp
```diff
@@ -55,3 +55,12 @@ DLL void* request_cuda_workspace(size_t nbytes, bool require_clean) {
...
     }
 }
+
+DLL void set_nccl_comms(void** comms, int num_comms) {
+    CudaContext::global()->nccl_comms = comms;
+    CudaContext::global()->num_comms = num_comms;
+}
+
+DLL void* get_nccl_comm(int idx) {
+    return CudaContext::global()->nccl_comms[idx];
```
yaoyaoding, line 65: Better to add some check logic here to make sure `idx < num_comms`.

yaoyaoding (commented):
Thanks @soodoshll !

Overall LGTM. Left some minor suggestions. We can merge this first and then work on other primitives and graph level operators.

PR #282, opened 2023-06-18
[External Tasks] Move task build into method call for external kernel support

xinli-git:
This is to support adding tasks from outside of the hidet library that uses external build logic. 

User can directly overwrite the `build` function to insert custom logic for introducing operators to hidet graph


yaoyaoding:
Thanks @xinli-git !

PR #285, opened 2023-06-20
[Operators] Conv2d gemm fp16 performance enhancement

Aalanli:
for pointwise convolution, use matmul.


Aalanli:
Additionally, I removed an extra cp_async statement in fp16_matmul

PR #288, opened 2023-06-22
[Document] fix installation guide

soodoshll:
Merely assigning environment variables is insufficient for setting up dev environment now. We need to run pip to install hidet package in develop mode.

Users still need to build source files written in C++ manually. Consider integrating that into `setup.py` in the future?


yaoyaoding:
Thanks @soodoshll !

PR #289, opened 2023-06-22
[FixBug] Fix bug where is_exit returns the opposite value

xinli-git:
_is_exit is set to True by at_exit, and False in the usual case, the logic currently is flipped.

Also, I don't think the None check is needed here. deleting it as well unless I missed something

### DIFF
--- a/python/hidet/cuda/stream.py
+++ b/python/hidet/cuda/stream.py
@@ -182,10 +182,9 @@ class StreamContext:
         from hidet.ffi import runtime_api
 
         current_streams = _current_streams
-        if self.device in current_streams:
-            self.prev_stream = current_streams[self.device]
-        else:
-            self.prev_stream = None
+        if self.device not in current_streams:
+            current_streams[self.device] = default_stream(self.device)
+        self.prev_stream = current_streams[self.device]
         current_streams[self.device] = self.stream
         self.device_context.__enter__()
         runtime_api.set_current_stream(self.stream)
@@ -194,9 +193,8 @@ class StreamContext:
         from hidet.ffi import runtime_api
 
         current_streams = _current_streams
-        if self.prev_stream is not None:
-            current_streams[self.device] = self.prev_stream
-            runtime_api.set_current_stream(self.prev_stream)
+        current_streams[self.device] = self.prev_stream
+        runtime_api.set_current_stream(self.prev_stream)
         self.device_context.__exit__(exc_type, exc_val, exc_tb)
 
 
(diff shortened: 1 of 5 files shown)


yaoyaoding:
Hi @xinli-git,

The check for None is added when I found that the exiting module can be freed when user call the is_exiting() function. Thus, `_is_exiting` is None also indicates the interpreter is exiting.

xinli-git:
Thnaks!, I added it back

python/hidet/runtime/storage.py
```diff
@@ -54,9 +54,11 @@ def memory_info(self) -> (int, int):
 
 class CudaMemoryAPI(MemoryAPI):
     def malloc(self, nbytes: int) -> int:
+        if nbytes == 0:
+            return 0
         with hidet.cuda.device(self.device.id):
-            addr = hidet.cuda.malloc_async(nbytes)
-        if addr == 0 and nbytes != 0:
+            addr = hidet.cuda.malloc(nbytes)
+        if addr == 0:
```
yaoyaoding, line 61: When we call `cudaMalloc` with `nbytes==0`, it will return 0. Thus, we also need to check `nbytes != 0` to make sure it is indeed out-of-memory. The `nbytes==0` will happen when we use tensor like `hidet.randn(shape=[0])`.

python/hidet/cuda/stream.py
```diff
@@ -191,8 +194,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         from hidet.ffi import runtime_api
 
         current_streams = _current_streams
-        current_streams[self.device] = self.prev_stream
-        runtime_api.set_current_stream(self.prev_stream)
+        if self.prev_stream is not None:
+            current_streams[self.device] = self.prev_stream
+            runtime_api.set_current_stream(self.prev_stream)
```
yaoyaoding, line 199: Thanks for noticing this bug. By default, all devices are on the "default stream" (in hidet, it is `default_stream(device=device_id)`). The updated implementation will not reset to the default stream when user exits the StreamContext. 

Consider implement in this way:
```
if self.device not in current_streams:
    current_streams[self.device] = default_stream(self.device)
...
```
and keep the original version.

xinli-git: thanks!

python/hidet/runtime/storage.py
yaoyaoding: I see, then it might be the feature of cudaMallocAsync.

I did not see the new added check for nbytes==0. Then it looks good for me.

xinli-git:
This fix is wrong, I will create a new one.

PR #291, opened 2023-06-24
[FixBug] Don't instantiate symbol for primitive functions

hjjq:
Previously, if a primitive function calls a primitive function, the `instantiate_symbols` pass will update the corresponding `hidet.ir.primitives.func.PrimitiveFunctionRegistry.function` in-place (I am not sure exactly how it's done, but this is what I observed), adding symbol variables to its parameters. The primitive function pool is a global variable, therefore this effect is cumulative across tuning candidates. So while candidate 0 will have no problem, candidate 1 will have two extra copies of symbol params, and so on, leading to compile errors.

Since primitive functions do not need symbol vars, a quick fix is just to not instantiate any symbols for them.


yaoyaoding:
Thanks @hjjq.

PR #292, opened 2023-06-26
[Tools] simple benchmarking utility

Aalanli:
See example.py for usage.


python/hidet/utils/benchmark/bench.py
```diff
@@ -0,0 +1,221 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import List, Optional, Callable, Tuple, Any, Dict, Union
+import time
+from dataclasses import dataclass
+
+import numpy as np
+import torch
+from matplotlib import pyplot as plt
```
yaoyaoding, line 18: I am trying to avoid the direct dependency of hidet on pytorch, because there might be other users of hidet (like people use onnx).

Now, hidet will directly import `hidet.utils`, thus, when people import hidet, the torch would be imported.
Consider moving the `import torch` and `import matplotlib` inside the function that used it. So that people do not need to install pytorch to `import hidet`.

python/hidet/utils/benchmark/example.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+bn = Bench(x_name='C', x_vals=[128 * i for i in range(2, 4)], config=BenchMatmulConfig())
+bn.bench(torch_matmul)
+bn.bench(hidet_matmul)
+bn.measure_flops(lambda config, c: torch.finfo(config.dtype).bits // 8 * c**2)
+
+data = bn.run()
+data.show_plot()
+data.print_data()
```
yaoyaoding, line 56: Let's move this to `hidet/tests/utils/test_benchmark.py` to
1. make the codebase more clean (the usage code can be put at other places).
2. check the correctness in ci after every commit to prevent the usage become stale.

yaoyaoding:
Thanks @Aalanli !

PR #293, opened 2023-06-27
[Fixbug] Is exiting fix

xinli-git:
_is_exiting is set to True by at_exit, and False in the usual case, the logic currently is flipped.

Now that is_exiting is working properly, we are actually clearning up memory. As a result, I found a different bug where dummy input creation is not properly handled for different data types (e.g. int types that are used for indexing)

This was not found before because we are re-using old memory regions and somehow these old allocations had proper values in them.


xinli-git:
The changes seems to pass, will merge. CC @yaoyaoding

PR #298, opened 2023-06-30
[Operators] preliminary symmetric weight quantization

Aalanli:
- Added preliminary symmetric weight quantization in tasks
- Changed Operator._run to Operator.symbolic_run, such that construction of the operator does not automatically run imperative_run, but symbolic_run. This is so that constant folding does not happen with barrier op, and leads to the intended behaviour with quantization layers. This should be fine to do since all imperative calls calls op.get_output(...), anyways, which calls imperative_run


python/hidet/graph/ops/quant/symmetric.py
```diff
@@ -0,0 +1,77 @@
...
+            name='symmetric_quantization_task',
+            inputs=[w],
+            outputs=[wq, scale],
+            attributes={'dims': dims, 'quant_type': quant_type},
+        )
+
+
+class SymmetricDeQuantizationTask(Task):
+    def __init__(self, wq: TensorNode, scale: TensorNode, dims=[-1]):
+        dims = [i if i >= 0 else len(wq.shape) + i for i in dims]
```
yaoyaoding, line 41: 1. Better to avoid use list as the default parameter. It is not recommanded by python because list is mutable. And in some IDE (like PyCharm), it will be highlighted and looks bad.
2. Consider using the `hidet.graph.ops.utils.normalize_dim` to normalize the dimensions. To be consistent with other hidet operators, it is better to allow the dims to be a single integer or a sequence of integers (e.g., `dims: Union[int, Sequence[int]]`).

python/hidet/graph/ops/quant/symmetric.py
```diff
@@ -0,0 +1,77 @@
...
+        self._assert(all(i >= 0 or i < len(wq.shape) for i in dims), "dims are out of bounds")
+
+        def unscale_weight(*indices):
+            scale_indices = [indices[i] for i in range(len(indices)) if not i in dims]
+            return cast(wq[indices], scale.type.dtype) / scale[scale_indices]
+
+        w = compute(name='dequantize', shape=wq.shape, fcompute=unscale_weight)
+        super().__init__(
+            name='symmetric_dequantization_task', inputs=[wq, scale], outputs=[w], attributes={'dims': dims}
```
yaoyaoding: To be consistent with other tasks, we usually do not write the `_task` suffix to the task name. It name will be printed when compilation, and `_task` will looks redundent.

python/hidet/ir/compute/cops/reduce.py
```diff
@@ -0,0 +1,56 @@
...
+    for i in range(len(x.shape)):
+        if i in dims:
+            if keep_dim:
+                y_shape.append(1)
+        else:
+            y_shape.append(x.shape[i])
+    return y_shape
+
+
+def reduce_cop(x: TensorNode, dims: List[int], keep_dim: bool, reduce_type: str, accumulate_dtype: str = 'float32'):
```
yaoyaoding: We do not need to add the `_cop` suffix to the compute operators as we will always place them in the `cops` namespace. Thus, when we use them, we can code like
```
from hidet.ir.compute import cops
cops.reduce(...)
cops.matmul(...)
```
without worrying the name conflicts (say with the ones under hidet.ops).

python/hidet/graph/nn/linear.py
```diff
@@ -61,3 +62,23 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+class SymQuantLinearTransposed(Module):
+    def __init__(self, weight: Tensor, bias: Optional[Tensor] = None, quant_type: str = 'int8'):
+        super().__init__()
+        self.in_features = weight.shape[0]
+        self.out_features = weight.shape[1]
+        qweight, scale = ops.symmetric_quantize(weight, quant_type=quant_type, dims=[-1])
+        self.qweight = qweight
+        self.scale = scale
+        self.bias = bias
```
yaoyaoding, line 74: This operator looks very different from the normal operator (e.g., only pass the operator config to the __init__ and create the tensor inside __init__). Is this nn module necessary? I can hardly think of a way to use it in hidet because the quantized operator is usually added in a pass.

yaoyaoding (changes requested):
Thanks @Aalanli !

I left some comments, and made some update directly on this PR.

I refactored a little bit on the `run`, `symbolic_run` and `imperative_run`.

Aalanli (commented):
Thanks for the changes, makes sense.

yaoyaoding:
Thanks @Aalanli !

yaoyaoding (approved):

PR #302, opened 2023-07-03
[Enhancements] add a vcude device to help mitigate compile time GPU memory usage

xinli-git:
## The problem
During compilation, some passes needs to replace certain operators with one or more new operators. For example, resolve_variants, subgraph_rewrite. In doing so, hidet relies on `imperative_run` to generate the new operators. This creates a lot of intermediate tensors that can exceed the actual run-time memory consumption, which makes larger models (such as Llama-7B) unable to compile even for a GPU with 24GB using fp16. 

## Fix
We introduce a vcuda device that allows GPU tensors to be stored on CPU and only transferred to GPU on demand. We call this v`cuda`. With this change, any additional GPU memory usage during compilation is off-loaded to CPU.

This might incur a bit compilation overhead when enabled, but given the time-consuming nature of such large models, this compile time increase is negligible. 

Now, on RTX3090, this is the GPU memory consumption for running llama test

```
Status of cuda:0 memory pool
   Allocated: 14081 MiB
        Peak: 14081 MiB
    Reserved: 1196 MiB
      Active: 12884 MiB
Status of cpu memory pool
   Allocated: 4 KiB
        Peak: 26022 MiB
    Reserved: 4 KiB
      Active: 0 Bytes
Status of vcuda:0 memory pool
   Allocated: 0 Bytes
        Peak: 25486 MiB
    Reserved: 0 Bytes
      Active: 0 Bytes
```

### DIFF
--- a/python/hidet/graph/flow_graph.py
+++ b/python/hidet/graph/flow_graph.py
@@ -528,6 +528,50 @@ class FlowGraph:
 
         return free_vars, nodes, usage_count
 
+    def _to_vcuda(self) -> None:
+        """
+        casts the flow graph object to vcuda device in place
+        """
+        from hidet.runtime.device import instantiate_device, Device
+
+        for x in self.inputs:
+            if not x.device.is_cuda():
+                raise ValueError("Inputs must be on cuda device")
+            x._move_to_vcuda()
+
+        for node in self.nodes:
+            if 'device' in node.attrs:
+                dev = instantiate_device(node.attrs['device'])
+                if dev.is_cuda():
+                    dev = Device('vcuda', dev.id)
+                node.attrs['device'] = dev
+            for inp in node.inputs:
+                inp._move_to_vcuda()
+            for outp in node.outputs:
+                outp._move_to_vcuda()
+
+    def _from_vcuda(self) -> None:
+        """
+        casts the flow graph object from vcuda device in place
+        """
+        from hidet.runtime.device import instantiate_device, Device
+
+        for x in self.inputs:
+            if not x.device.is_vcuda():
+                raise ValueError("Inputs must be on vcuda device")
+            x._move_from_vcuda()
+
+        for node in self.nodes:
+            if 'device' in node.attrs:
+                dev = instantiate_device(node.attrs['device'])
+                if dev.is_vcuda():
+                    dev = Device('cuda', dev.id)
+                node.attrs['device'] = dev
+            for inp in node.inputs:
+                inp._move_from_vcuda()
+            for outp in node.outputs:
+                outp._move_from_vcuda()
+
 
 def trace_from(tensor: Union[Tensor, List[Tensor]], inputs: Optional[Union[Tensor, List[Tensor]]] = None) -> FlowGraph:
     """
--- a/python/hidet/graph/flow_graph.py
+++ b/python/hidet/graph/flow_graph.py
@@ -537,7 +537,7 @@ class FlowGraph:
         for x in self.inputs:
             if not x.device.is_cuda():
                 raise ValueError("Inputs must be on cuda device")
-            x._move_to_vcuda()
+            x.move_to_vcuda()
 
         for node in self.nodes:
             if 'device' in node.attrs:
@@ -546,9 +546,9 @@ class FlowGraph:
                     dev = Device('vcuda', dev.id)
                 node.attrs['device'] = dev
             for inp in node.inputs:
-                inp._move_to_vcuda()
+                inp.move_to_vcuda()
             for outp in node.outputs:
-                outp._move_to_vcuda()
+                outp.move_to_vcuda()
 
     def from_vcuda(self) -> None:
         """
@@ -559,7 +559,7 @@ class FlowGraph:
         for x in self.inputs:
             if not x.device.is_vcuda():
                 raise ValueError("Inputs must be on vcuda device")
-            x._move_from_vcuda()
+            x.move_from_vcuda()
 
         for node in self.nodes:
             if 'device' in node.attrs:
@@ -568,9 +568,9 @@ class FlowGraph:
                     dev = Device('cuda', dev.id)
                 node.attrs['device'] = dev
             for inp in node.inputs:
-                inp._move_from_vcuda()
+                inp.move_from_vcuda()
             for outp in node.outputs:
-                outp._move_from_vcuda()
+                outp.move_from_vcuda()
 
 
 def trace_from(tensor: Union[Tensor, List[Tensor]], inputs: Optional[Union[Tensor, List[Tensor]]] = None) -> FlowGraph:
--- a/python/hidet/graph/flow_graph.py
+++ b/python/hidet/graph/flow_graph.py
@@ -528,7 +528,7 @@ class FlowGraph:
 
         return free_vars, nodes, usage_count
 
-    def to_vcuda(self) -> None:
+    def vcuda_(self) -> None:
         """
         casts the flow graph object to vcuda device in place
         """
@@ -537,7 +537,7 @@ class FlowGraph:
         for x in self.inputs:
             if not x.device.is_cuda():
                 raise ValueError("Inputs must be on cuda device")
-            x.move_to_vcuda()
+            x.vcuda_()
 
         for node in self.nodes:
             if 'device' in node.attrs:
@@ -546,11 +546,11 @@ class FlowGraph:
                     dev = Device('vcuda', dev.id)
                 node.attrs['device'] = dev
             for inp in node.inputs:
-                inp.move_to_vcuda()
+                inp.vcuda_()
             for outp in node.outputs:
-                outp.move_to_vcuda()
+                outp.vcuda_()
 
-    def from_vcuda(self) -> None:
+    def cuda_(self) -> None:
         """
         casts the flow graph object from vcuda device in place
         """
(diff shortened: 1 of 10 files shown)


python/hidet/graph/operator.py
```diff
@@ -84,6 +84,8 @@ def build_target(self) -> str:
 
         if isinstance(self, TransferOp):
             return 'cuda'
+        elif self.device.is_vcuda():
+            return 'cuda'
         else:
             return self.device.kind
```
yaoyaoding, line 90:
```suggestion
        else:
            if self.device.kind in ["cuda", "vcuda"]:
                return "cuda"
            elif self.device.kind == "cpu":
                return "cpu"
            else:
                raise NotImplementedError()
```
Hope this to be more readable.

python/hidet/graph/tensor.py
```diff
@@ -789,6 +789,45 @@ def cuda(self, device=None):
...
+        """
+        if self.device.is_cuda():
+            return
+        if not self.device.is_vcuda():
+            raise ValueError("Tensor must be on vcuda device, got {}".format(self.device))
+
+        if self.storage is not None:
+            self._storage = self.storage.cuda(self.device.id)
+        self._device = Device('cuda', self.device.id)
+
```
yaoyaoding, line 830: Similar to the graph. Use "vcuda_()" and "cuda_()".

python/hidet/runtime/compiled_task.py
```diff
@@ -176,8 +176,21 @@ def run_async(self, inputs):
...
+
         candidate = self.candidates[self.pick_best_candidate(inputs, outputs)]
         candidate(*inputs, *outputs)
+
+        if has_vcuda:
+            for inp in inputs:
+                inp.move_to_vcuda()
+            for outp in outputs:
+                outp.move_to_vcuda()
+
```
yaoyaoding, line 193: There are two ways to implement "vcuda" device:
1. [The current version]: we copy the tensor to cuda before we run the operator and copy back to cpu when we finish it.
2. We can also reply on the unified addressing of cuda. (see [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#unified-virtual-address-space) and [here](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html)). Essentially, the unified addressing allows the cuda kernel access the host memory (page-locked) directly. Thus, we do not need to copy to and from the cuda device memory.

I prefer the second way. With it, we do not need to update our `run_async` function.

python/hidet/runtime/compiled_task.py
```diff
@@ -213,7 +226,10 @@ def _check_inputs(traced_inputs: Iterable[TensorSignature], inputs):
 
     symbol_map = {}
     for i, (traced, new) in enumerate(zip(traced_inputs, inputs)):
-        if traced.device.partition(':')[0] != new.device.kind:
+        traced_dev_kind = traced.device.partition(':')[0]
+        if traced_dev_kind != new.device.kind:
+            if traced_dev_kind == 'cuda' and new.device.is_vcuda():
+                continue
```
yaoyaoding, line 232: Let's add a `target` attribute to the Device class? 

Like 
```python
class Device:
    kind: str  # 'cpu', 'cuda', 'vcuda'
    target: str  # 'cpu', 'cuda'
    id: int  # 0, 1, ...
```
With it, we do not need the special handing here. It will also make the `FlowGraph.build_target` slightly simpler.

python/hidet/testing/models/llama.py
```diff
@@ -357,8 +357,8 @@ def convert_model(hf_model: torch.nn.Module, dtype=hidet.float16, device='cuda')
 
 def build_flow_graph(model, batch_size=1, device='cuda', dtype='float16'):
     config = model.config
-    input_ids = hidet.symbol([batch_size, "seq_length"], dtype=hidet.int32, device=device)
-    position_ids = hidet.symbol([batch_size, config.max_position_embeddings], dtype=hidet.int32, device=device)
+    input_ids = hidet.symbol([batch_size, "seq_length"], dtype=hidet.int64, device=device)
```
yaoyaoding, line 360: I know there is a error to use "hidet.int32" here, but is there a way to fix that error directly instead of using "int64"?

python/hidet/testing/models/llama.py
```diff
@@ -438,8 +438,7 @@ def generate_torch(input_ids: str, tokenizer, torch_model, num_tokens, device='c
 def get_compiled_model(name='decapoda-research/llama-7b-hf', device='cuda', opt=False):
     tok = LlamaTokenizer.from_pretrained(name)
 
-    with torch.device("cuda"):  # reduce the time to load the model
-        model = hfLm.from_pretrained(name, torch_dtype=torch.float16)
+    model = hfLm.from_pretrained(name, torch_dtype=torch.float16)
```
yaoyaoding, line 441: Use the 
```python
with torch.device("cuda"):
   ...

...
torch.cuda.empty_cache()
```
could greatly reduce the weight initialization time. Does it increate the overall system memory usage?

python/hidet/testing/models/llama.py
```diff
@@ -450,7 +449,9 @@ def get_compiled_model(name='decapoda-research/llama-7b-hf', device='cuda', opt=
     flow_graph = build_flow_graph(model, device=device)
 
     if opt:
-        flow_graph = hidet.graph.optimize(flow_graph)
+        with hidet.graph.PassContext() as ctx:
+            ctx.reduce_cuda_compile_mem(True)
```
yaoyaoding, line 453: We use some heristic rule to determine whether to turn on this.

For example, if the model weights used more than 1/3 gpu memory to store its weights, we can apply the transformation, otherwise disable it.

```python
def reduce_cuda_compile_mem(flag: Optional[bool] = None):
    """
    flag: Optional[bool]
        None - determined automatically
        True - always ...
        False - never ...
    """
    ...
```
We add the instrument when the `flag is None or flag is True`and do the checking in the instrument.

yaoyaoding (commented):
Hi @xinli-git , good work!

I left some alternative implementation suggestions and let's discuss.

python/hidet/graph/flow_graph.py
```diff
@@ -528,6 +528,50 @@ def find_all_nodes(u: Operator):
...
+            if 'device' in node.attrs:
+                dev = instantiate_device(node.attrs['device'])
+                if dev.is_vcuda():
+                    dev = Device('cuda', dev.id)
+                node.attrs['device'] = dev
+            for inp in node.inputs:
+                inp.move_from_vcuda()
+            for outp in node.outputs:
+                outp.move_from_vcuda()
+
```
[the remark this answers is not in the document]
yaoyaoding: We can define the semantics of the device conversion as the combination of the following two steps:
1. convert all leaf nodes (i.e., the graph inputs and graph weights, or in another word, all the reachable tensors with `tensor.trace == None`) to the target device.
2. reforward all the nodes in the flow graph in order.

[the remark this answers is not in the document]
yaoyaoding: We raise an `Error` if there are more than one device for the graph inputs and weights.

python/hidet/testing/models/llama.py
xinli-git: deleted by mistake, thanks for spotting

python/hidet/testing/models/llama.py
xinli-git: fixed by casting the model output to int32

python/hidet/runtime/compiled_task.py
xinli-git: this is a great suggestion and it works! thank you

python/hidet/graph/flow_graph.py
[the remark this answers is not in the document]
xinli-git: you do mean to introduce a new functor to do this?

python/hidet/testing/models/llama.py
yaoyaoding: I see. From https://github.com/hidet-org/hidet/blob/f937ad23ef28558d205276f7588cca805377370a/python/hidet/graph/ops/reduce/reduce.py#L209, we seems force the output of arg reduce to `int64` (seems to be consistent with pytorch, #111). 

Might be better to add a parameter to all arg reduce operator (e.g., `def argmin(..., out_dtype='int64')`) to allow we use other data types.

python/hidet/runtime/compiled_task.py
yaoyaoding: One minor issue is that the cpu tensor allocated by other library (e.g., pytorch and numpy) will not use `cudaMallocHost` to allocate, and it is not page-locked. Thus, when we convert the tensors from them, we should allocate the tensor by hidet instead of sharing the weights (on the platform that cuda is avaliable).

python/hidet/graph/flow_graph.py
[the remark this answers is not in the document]
yaoyaoding: Em... I think this problem again, and find that it is better to keep the original design.

The semantics of `to_vcuda` and `from_vcuda` is convert all `cuda` tensor in the flow graph to `vcuda` and vise versa.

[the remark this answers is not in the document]
yaoyaoding: In this semantics, we should use a "if" make sure we only change the "cuda" tensor and leave other "cpu" tneosr untouched.

python/hidet/runtime/compiled_task.py
xinli-git: I think this case is fine because if users give a cpu tensor from other library and expect it to work on CUDA device that is already wrong

xinli-git: runtime check should catch this

yaoyaoding: Make sense.

yaoyaoding (approved):
Thanks @xinli-git !

PR #311, opened 2023-07-12
[Reduce] optimize and unify reduce operator to a single place

xinli-git:
This changes optimizes the reduce operator and unifies both reduce and reduce_f16 into a single class / implementation.

On Llama, 255 tokens (RTX3090)
Before: 6.173s
After: 5.295s

Contribution from reduce:
before: 732ms (13.6% of total)
after: 126ms (2.7% of total)


python/hidet/graph/ops/reduce/resolve.py
```diff
@@ -35,25 +34,32 @@ def resolve_simplify(self, op: Operator) -> Optional[List[Tensor]]:
...
+            # for some key models, the reduction dimension spans over multiple dims
+            # e.g. 40 x 32 x 32. In this case, it is best to map the reduction as
+            # a 2-D tensor so the warp reduction implementation does not need to
+            # handle this special case and fusion can still work
+            out_shape = op.outputs[0].shape
+            spatial = prod(shape[i] for i in range(len(shape)) if i not in dims)
+            reduce = prod(shape[i] for i in dims)
+            x = x.reshape((spatial, reduce))
+            op_cls = op.__class__
+            x = op_cls(x, [-1], keepdims).outputs[0]
```
yaoyaoding, line 48: Can we use `reforward` here?

python/hidet/graph/ops/reduce/reduce.py
```diff
@@ -48,69 +53,110 @@ def allow_epilogue(self) -> bool:
...
-        block_size = warp_size
         x, y = self.inputs[0], self.outputs[0]
+        warp_size = 32
         xdtype = x.type.dtype
         shape: List[Int] = list(x.shape)
+        lanes = 1
+        vtype: DataType = xdtype
+        if xdtype.nbytes < 4:
+            num_eles: int = 4 // xdtype.nbytes
+            if shape[-1] % num_eles == 0:
```
yaoyaoding, line 85: consider adding an `assert is_constant(shape[-1])` if this schedule template only supports the case where the last dimention is known.

yaoyaoding: Or change it to
```python
if is_constant(shape[-1]) and shape[-1] % num_eles == 0:
    ...
```

yaoyaoding (approved):
Thanks @xinli-git ! You are getting more professional in using hidet script :)

PR #313, opened 2023-07-12
[Graph] Minor graph benchmark fix

Aalanli:


python/hidet/graph/graph_utils/instruments/benchmark_instrument.py
```diff
@@ -57,10 +57,17 @@ def after_operator(self, op: Operator, inputs: List[Tensor], outputs: List[Tenso
...
 
-        task_func: CompiledModule = op.compiled_task
-        latency: List[float] = task_func.profile(
-            *inputs, *outputs, warmup=self.warmup, number=self.number, repeat=self.repeat
-        )
+        task_func = op.compiled_task
+        if isinstance(task_func, CompiledModule):
+            latency: List[float] = task_func.profile(
+                *inputs, *outputs, warmup=self.warmup, number=self.number, repeat=self.repeat
+            )
```
yaoyaoding, line 64: Hi @Aalanli, can I know in which case the `op.compiled_task` will be an `CompiledModule`. By our design, it should always be `CompiledTask`, even it only contains a single `CompiledModule` inside it.

Aalanli: Right, I wasn't sure, so I left that there. If the intention is CompiledTask, then I can remove CompiledModule branch.

yaoyaoding: Yeah, it will always returns `CompiledTask`. It is safe to reply on the type hint in hidet.

PR #315, opened 2023-07-13
[Task] More descriptive kernel names for nsys/ncu

Aalanli:


python/hidet/ir/schedulers/cpu/scheduler.py
```diff
@@ -26,7 +26,12 @@ def schedule_grid_compute(self, node: GridCompute, tensor_map: Dict[TensorNode,
 
         params, param_map, call_args = self.grid_compute_params_and_args(node, tensor_map)
 
-        with FunctionBuilder(name=f'compute_{node.name}', kind='cpu_kernel') as fb:
+        if self.task is not None:
+            name = f'{self.task.name}_compute_{node.name}'
+        else:
+            name = f'compute_{node.name}'
+
+        with FunctionBuilder(name=name, kind='cpu_kernel') as fb:
```
yaoyaoding, line 34: When the auto scheduler is scheduling a task, the `self.task` will be the task that is scheduling, thus we do not need to add the `if` here.

yaoyaoding (approved):
Thanks @Aalanli, this would be helpful when we use nsight system and compute to profile the kernels.

PR #316, opened 2023-07-16
[Operator] optimize normalize op with vectorized load, dynamic shape and more

xinli-git:
this change introduces several enhancements to the current norm operator

* vectorized load for fp16 types
* allow epilogue
* dynamic shape on the normalized dimension 
* add a tuning to use 2 warp shuffle routines or just a single one
* cleaner code and implementation

as a result, norm_fp16.py can be safely deleted

for shapes in stable diffusion: [2, 32, 60, 16, 16], norm dims [60, 16, 16], fp32:
* torch : 0.023 ms
* main: 0.036 ms
* this change: 0.027 ms

for shapes in bert-base: [1, 128, 768], norm dims [768], fp16:
* torch: 0.006 ms
* main: 0.008 ms 
* this change: 0.007 ms

We are still not faster than torch, but very close


xinli-git:
the tests are passing 
this change is isolated to norm OP so will merge without a review.

python/hidet/graph/ops/normalize/norm.py
```diff
@@ -104,39 +108,69 @@ def allow_prologue(self) -> bool:
         return False
 
     def allow_epilogue(self) -> bool:
```
czhxiaohuihui, line 110: 问一个比较小白的问题，prologue 和这里的 epilogue，在中文里面一般怎么翻译，二者是什么作用呢，谢谢

PR #318, opened 2023-07-17
[Bugfix] Fix associated bugs with dynamic shape tracing

Aalanli:
When `torch.compile(model, backend="hidet", dynamic=True)`, sometimes there are `SymInt`s in the input, these could be either symbolic or concrete.

```python
def forward(self, tensor : torch.Tensor, seq_len : torch.SymInt):
    view = tensor.view(1, seq_len, 16, 64);  tensor = seq_len = None
    transpose = view.transpose(1, 2);  view = None
    contiguous = transpose.contiguous();  transpose = None
    return (contiguous,)
```

When actually run, the associated inputs are torch tensors on cpu with shape `[]`. These are treated like python scalars in torch. Hence, the changes are
1. Make `hidet.Tensor` behave more like a python scalar if shape is `[]` and device is cpu
2. Enable `SymbolVar` to be passed as input to the flowgraph. Is is to cover cases that 0-dimensional cpu Tensors can't handle, like passing parameter to reshape, or constructing a tensor with dynamic shape. Or indeed any function that expects a python scalar.
3. Fix a bug in `AddExplicitCastPass` so that the following script works:

```python
import hidet
k = hidet.symbol(['a'])
a = k.shape[0]
ad = a + 1
b = hidet.arange(a, ad, dtype='int64', device='cuda')
w = hidet.randn([1024, 1024], device='cuda')
y = hidet.ops.take(w, b, axis=0)

g = hidet.trace_from([y], inputs=[a])
g = hidet.graph.optimize(g)

cg = g.build()
y = cg(hidet.full([3], 2, dtype='int32'))
```

Originally, this script won't even compile.

I think its a problem in `Namer`, where two `Var`s are equal iff they are the same object, not if they have the same `id` and `type`, so that any pass that copies or constructs a new `Var` with identical id and hint `c` would not be in dict `self.obj_name` in name, and consequently gets renamed to `c_1`.

The following script now passes without any errors.
```python
import torch
from transformers import PegasusForConditionalGeneration, PegasusTokenizer
import hidet
src_text = [""" The tower is """]

tokenizer = PegasusTokenizer.from_pretrained("google/pegasus-cnn_dailymail")

model = PegasusForConditionalGeneration.from_pretrained("google/pegasus-cnn_dailymail").eval().cuda()
model.model.encoder = torch.compile(model.model.encoder, backend='hidet', dynamic=True)
model.model.decoder = torch.compile(model.model.decoder, backend='hidet', dynamic=True)

batch = tokenizer(src_text, truncation=True, padding="longest", return_tensors="pt").to('cuda')
translated = model.generate(**batch, max_length=32, num_beams=1, early_stopping=True)
tgt_text = tokenizer.batch_decode(translated, skip_special_tokens=True)
print(tgt_text)
```


yaoyaoding:
Hi @Aalanli, thanks for your work! But, after some thinking, I decide not take the design to add symbolic scalar as the input of a flow graph. 

Here are some of my considerations:
1. It is important to keep the IR as simple and intuitive as possible, otherwise the system is easy to become too complex to maintain. Adding the scalar as the graph inputs is not necessary for most use cases (i.e., we do not encounter any model that requires this). For the torch dynamo (dynamic=True) case, we can implement the scalar inputs in the torch dynamo frontend of hidet, and it is not necessary to change our IR. See #319 .
2. It is not consistent to only add scalar as the graph input. Do we need to add scalar output for the flow graph? Do we need to add scalar input/output for the operators? 

We can talk offline if you want to discuss. And it is always better to have some chat with me before you want to do some change to the important IRs.

PR #325, opened 2023-07-21
[Fixbug] Reduce should perform syncthread after initializing shared memory to zero

xinli-git:
This bug slipped through the CI because the test shapes were all too small. There is an initialization of shared memory that is not guarded by syncthread before others were able to access it.


yaoyaoding:
Thanks @xinli-git !

PR #333, opened 2023-07-27
[Models] Llama2 fix

Aalanli:
Fix llama2 and add test for num_heads != num_key_value_heads.


Aalanli:
This is quite strange, it seems that while the tests I added works on my local machine, it does not work on the CI server. I am investigating if the transformer versions are the same or not.

yaoyaoding:
Have added a CI step to print out the version of installed python packages at #334 .

Aalanli:
Right, it seems that the transformer version is 4.29.2, when the latest one is 4.31.0.

yaoyaoding:
I see. The github ci will cache the python package (so that we do not need to reinstall the large packages like pytorch for every run), and this also makes some packages not up to date. 

Just cleared the python cache manually on the ci server.

Aalanli:
It appears that updating transformers broke torch dynamo for bert, specifically for dynamic=True. This has nothing to do with our backend, though.

yaoyaoding:
I see, could you skip the corresponding tests in this PR? Let's wait for the fix from pytorch people.

Issue #335, opened 2023-07-28
[RFC][Discussion] Automatic Parallelization


soodoshll:
[rendered rfc](https://github.com/soodoshll/hidet/blob/auto-parallel-rfc/rfcs/0003-auto-parallelization.md)

yaoyaoding:
Hi @soodoshll, thanks for the draft! 

It looks good as a first verion rfc draft!

I have several suggestions:
1. The 0001 and 0002 rfc slot has been used. Might consider use 0003.
2. Might be better to give a concrete example of `distributed_config` and `optimization_config` in the guide-level explanation. 
3. Reference-level explanation is a good place to show what configs are available for `distributed_config` and `optimization_config`, and what each config specify. It's okay only to put the ones that are known now and update the draft and add more configs during implementation and refactor in the future.
4. I prefer putting the out_dir as a seperate function parameter instead of an attribute in config.
5. Add a reference to Alpa and use something like "Alpa[1]" in the main text.
6. This part "For example, for a 4x4 multi-machine-multi-GPU cluster, the possible sharding specifications are (4x4, 1), (4(machine), 4(gpus)), (4(gpus), 4(machines)), (1, 4x4). We do not consider (2, 8) or (8, 2). Therefore, using R or Si is sufficient since the number of shards is determined by the number of devices. " is a little vague. We can consider adding some example to illustrate what does a specific sharding specification mean (e.g., (4 gpus, 4 machine)), and explain the meaning of "R" and "Si". 
7. Consider using `mesh_axes_per_dim` in `TensorShardSpec`. 
8. The math formula in "Operator Sharding Specification" has some typesetting flaws.

yaoyaoding:
9. Could add a section to describe the ILP formulation.

The design looks good to me. Hi @soodoshll and @xinli-git, could you also discuss how to seperate the whole feature into relative small steps to implement? We can use this issue to track the PRs related to this RFC, something like https://github.com/apache/tvm/issues/15319. Thanks!

xinli-git:
Thanks! @soodoshll. The RFC is very detailed. 

For modelling computation, it seems that Alpa assumes that all tensor contraction OPs (MM, Conv) must be fully sharded so all such ops that same computation cost under different sharding strategies. They also observe that other OPs have negligible runtime cost for computation. (I verified this as well). As a result, they think there was no need to model computation. 

[...]

PR #336, opened 2023-07-31
[Distributed][auto-parallel] Sharding Specification and rule discovery

soodoshll:
This PR tries to implement step 1) and 3) mentioned in https://github.com/hidet-org/hidet/issues/335#issuecomment-1656568711

It contains an example of automatically discovering the sharding rules of all operators in resnet model, which involves dynamic batch size. You can run it with 

```python -m hidet.distributed.partition.rule``` 

Now we only support discovering 1D partitions.

The rule discovery process is basically to enumerate all possible input-sharding specifications and check their validity. The checking takes two steps:
1. Shape check: it will generate a sharded version of input tensors and re-forward the op to get the new output tensors. Then it will try to align the new output with the original whole output tensors, in order to find which output dimensions are sharded. For example, if the input $A$ is an 8x8 matrix, the op is $sin$, and we want to shard the first dimension of $A$ into 2 partitions, then it will generate an input shard with shape 4x8, and get the output shape 4x8, and we know that the first dimension of the output is also shared. 
2. Data dependency check. The validity of the shape check does not guarantee correctness. For example, if we have an 8x8 matrix $A$, and the op is $softmax$, then sharding along the second dimension gives the correct shape but wrong results, because the denominator of $softmax$ depends on the complete second dimension. So we analyze the range of data the output depends on, and the sharding is valid only if **the output shard only depends on the input shard**. For example, any shard of $A+1$ only depends on the same shard of $A$, so it is valid. We trace the indexing expressions(`TensorElement`) from `fcompute` of all outputs, to get the range of input indices that are accessed. And then we use Z3 SMT solver to check if any accessed input indices are out of the shard boundary. If not, the sharding is valid.

Note that even data dependency does not guarantee correctness completely (if the op plays tricks), but I find it can cover most of the ops I met. Though we still need to test it more thoroughly.

I can add more details about it to the RFC.

cc @yaoyaoding @xinli-git

### CODE REVIEW

**Comment by yaoyaoding** [comment]
Hi @soodoshll,

Can I know the main purpose of the z3 solver? 

Because we also have a bound analyzer as well as an arthimatic simplifier in hidet, is any limitation of our internal analyzer and simpilifer? 

We need to be cautious when introducing new dependency, thus more justification is needed when introducing a new one.

**Comment by yaoyaoding** [comment]
Make sense. For different sequence length, the best partition strategy might be different. Thus, we do need to optimize for some specific input size and then use the found stragegy for all sequence length.

**Comment by yaoyaoding** [comment]
Hi @soodoshll, ping me when the PR is ready to be reviewed. Thanks!

**Comment by yaoyaoding** [comment]
Hi @soodoshll, thanks for the update!

After you finish the whole pipeline, you can split it into several PRs and @xinli-git and I can help to review.

File: python/hidet/distributed/partition/rule.py [hunk 1]
```diff
@@ -0,0 +1,256 @@
...
+        cache = {}
+        for op in flow_graph.nodes:
+            print(op)
+            if str(op) not in cache:
+                shard_plans = op_shard_rule_search(op, 4)
+                cache[str(op)] = shard_plans
+            for sp in cache[str(op)]:
+                print(sp)
+
+    resnet_test()
```
yaoyaoding, line 256: Add some tests for the shard rule searching in `tests/distributed/partition/test_shard_rule.py`.

yaoyaoding (changes requested):
Thanks @soodoshll !

I left some comments. In general, it is highly suggested to add the type annotation for the class fields and function parameters so that we do not need to check the usage of the class/function to know what it expects. It will reduce the time needed to understand the code and review. We might need to write this to our developer guide.

The current discovery algorithm only tries to find the sharding rule for 1-d mesh. It is better to write a module-level comment to briefly describe what we have done, what's the assumption and what's the limitation.

python/hidet/distributed/partition/rule.py
```diff
@@ -0,0 +1,227 @@
...
+                if shard_dim >= 0:
+                    if inputs[i].shape[shard_dim] % num_shards != 0:
+                        break
+                    new_inputs.append(get_tile(inputs[i], shard_dim, num_shards))
+                else:
+                    new_inputs.append(inputs[i])
+            if len(new_inputs) < len(inputs):
+                continue
+            outputs = op.reforward(new_inputs)
+        except (ValueError, RuntimeError, AssertionError, IndexError):
```
xinli-git, line 188: how do we arrive at these exceptions? I assume through trial and error?

python/hidet/distributed/partition/shard.py
```diff
@@ -0,0 +1,237 @@
...
+        return self._sharded_dim == other._sharded_dim
+
+
+def get_tile(
+    tensor: Union[Tensor, TensorNode], shard_dim: int, nshards: int, rank: int = 0
+) -> Union[Tensor, TensorNode]:
+    # Only works for 1-D partition
+    slice_list = []
+    for i in range(len(tensor.shape)):
+        if i != shard_dim:
```
xinli-git, line 83: regarding the -1 vs None, I think the code would be cleaner with None meaning not sharded and avoid -1. For example, here,  we can just check if shard_dim is None and be more explicit.

xinli-git (commented):
I am still going through the code to understand the sharding rules better but posting some comments first. 

Feel free to merge without me once Yaoyao thinks it looks good :)

python/hidet/distributed/partition/rule.py
```diff
@@ -0,0 +1,227 @@
...
+    found_rules = []
+
+    # We do not allow dynamic shapes
+    assert len(op.task.symbols) == 0
+
+    # num_shards == 1 will break following logic
+    if num_shards == 1:
+        return [
+            OpShardSpec(
+                [TensorShardSpec(len(t.shape)) for t in op.inputs], [TensorShardSpec(len(t.shape)) for t in op.outputs]
```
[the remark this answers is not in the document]
xinli-git: a sorry, my bad

python/hidet/distributed/partition/rule.py
```diff
@@ -0,0 +1,227 @@
...
+from hidet.ir.functors import ExprRewriter, ComputeRewriter
+from hidet.ir.analyzers.bound_analyzer import infer_bound, BoundInfo
+
+
+from .shard import TensorShardSpec, OpShardSpec, get_tile, AllReduce, ReduceScatter
+
+
+class IndexRewriter(ExprRewriter, ComputeRewriter):
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
yaoyaoding: Hi @xinli-git , the current writing looks fine. This function does not require any explicit ordering of calling its parent classes' init method. The default one `super().__init__()` will call them in MRO order and it works as we expected. Explicitly calling some parent class's init method might ignore some other parent classes and is error prone.

python/hidet/distributed/partition/shard.py
```diff
@@ -0,0 +1,221 @@
...
+
+    The constructor can be called by specifying the dimension sharded by each mesh axis. The
+    parameter `sharded_dims` is a list of integers, and the i-th element of it is the dimension
+    sharded by the i-th mesh axis. If any mesh axis does not shard any dimension (tensor are
+    duplicated along that mesh axis), its corresponding value is None.  We assume that each mesh
+    axis can shard at most one dimension, and each dimension can be sharded by multiple mesh axes.
+
+    It also accept a single-value parameter, which will be viewed as a single-layer mesh hierarchy.
+
+    None or negative numbers will be treated as not to shard on this dimension
```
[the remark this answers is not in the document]
yaoyaoding: The latter one has better readability in this case to me :)

python/hidet/distributed/partition/rule.py
```diff
@@ -0,0 +1,227 @@
...
+    def __init__(self, op: Operator, num_shards: int):
+        self.op: Operator = op
+        self.num_shards: int = num_shards
+
+        index_rewriter = IndexRewriter()
+        for o in self.op.task.outputs:
+            if not isinstance(o, GridCompute):
+                self.valid = False
+                return  # we don't support scalar output
+            index_rewriter.visit(o)
```
yaoyaoding, line 75:
```suggestion
        index_rewriter = IndexRewriter()
        index_rewriter.visit(self.op.task.outputs)
        self.valid = all(isinstance(out, GridCompute) for out in self.op.task.outputs)
```

yaoyaoding (approved):
It looks good to me now. 

Feel free to merge after fixing the existing comments. Thanks @soodoshll !

PR #337, opened 2023-08-01
[Operator] Composite Elementwise Operation

hjjq:
Assuming `U1` and `U2` are unary elementwise operators, and `B` is a binary elementwise operator.
Introduce a new fused operator called Composite Elementwise Operation, where `comp_elemwise(x, U1, U2, B) = B(U1(x), U2(x))`.
Also add subgraph rewrite rule. This allows for more fusion opportunity.

### DIFF
--- a/python/hidet/graph/ops/arithmetic.py
+++ b/python/hidet/graph/ops/arithmetic.py
@@ -31,7 +31,7 @@ class UnaryElementwiseOperation:
         self.y: Expr = y
         self.attributes = attributes
         self.task_attributes = task_attributes
-        self.name = name
+        self.name: str = name
 
     def __call__(self, x):
         return rewrite(self.y, {self.x: x})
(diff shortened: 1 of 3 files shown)


python/hidet/graph/ops/arithmetic.py
```diff
@@ -15,11 +15,51 @@
...
+        y = op(x)
+        return UnaryElementwiseOperation(x, y, name, attributes, task_attributes)
+
+
+class BinaryElementwiseOperation:
+    def __init__(self, lhs, rhs, out, name):
+        self.lhs = lhs
+        self.rhs = rhs
+        self.out = out
+        self.name = name
```
yaoyaoding, line 51: Better to add some type hints to increase the readability.

yaoyaoding (approved):
Thanks @hjjq, it looks good to me!

PR #345, opened 2023-08-08
[Dynamo] minor enhancements to attention and register a few functions

xinli-git:
Encountered a few minor issues when compiling a transformer-based model using torch.compile with very large batch sizes, submitting the fix here.

### DIFF
--- a/python/hidet/graph/frontend/torch/register_functions.py
+++ b/python/hidet/graph/frontend/torch/register_functions.py
@@ -981,6 +981,28 @@ def torch_mean(
     return output
 
 
+@register_function(torch.sum)
+@register_method(torch.Tensor.sum)
+def torch_sum(x: Tensor, *, dtype: Optional[DataType] = None) -> Tensor:
+    if dtype:
+        x = x.astype(dtype_from_torch(dtype))
+    output = ops.sum(x, dims=list(range(len(x.shape))), keep_dim=True)
+    return output
+
+
+@register_function(torch.sum)
+@register_method(torch.Tensor.sum)
+def torch_sum(
+    x: Tensor, dim, keepdim=False, *, dtype: Optional[DataType] = None, out: Optional[Tensor] = None
+) -> Tensor:
+    if out is not None:
+        raise NotImplementedError("hidet: does not support torch.sum(..., out=...)")
+    if dtype:
+        x = x.astype(dtype_from_torch(dtype))
+    output = ops.sum(x, dims=dim, keep_dim=keepdim)
+    return output
+
+
 @register_function(torch.cumsum)
 def torch_cumsum(x: Tensor, dim, *, dtype: Optional[DataType] = None, out: Optional[Tensor] = None) -> Tensor:
     if out is not None:
--- a/python/hidet/graph/frontend/torch/register_methods.py
+++ b/python/hidet/graph/frontend/torch/register_methods.py
@@ -90,6 +90,11 @@ def tensor_to(self: Tensor, *args, **kwargs) -> Tensor:
             if self.is_symbolic() and instantiate_device(device_from_torch(arg)) != self.device:
                 raise NotImplementedError('hidet: Tensor.to(..., device=...) is not supported for symbolic tensors.')
             device = arg
+        elif isinstance(arg, Tensor):
+            dtype = arg.dtype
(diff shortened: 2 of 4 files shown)


python/hidet/graph/frontend/torch/register_functions.py
```diff
@@ -981,6 +981,28 @@ def torch_mean(
...
+    x: Tensor, dim, keepdim=False, *, dtype: Optional[DataType] = None, out: Optional[Tensor] = None
+) -> Tensor:
+    if out is not None:
+        raise NotImplementedError("hidet: does not support torch.sum(..., out=...)")
+    if dtype:
+        x = x.astype(dtype_from_torch(dtype))
+    output = ops.sum(x, dims=dim, keep_dim=keepdim)
+    return output
+
+
```
hjjq, line 1005: Why do we need two `torch_sum`s here?

hjjq (approved):
Thanks @xinli-git, LGTM. Just a minor question about torch_sum

python/hidet/graph/frontend/torch/register_functions.py
xinli-git: I followed the convention for the mean method above. Not entirely sure either as I thought python does not support overloading. Perhaps @yaoyaoding knows?

yaoyaoding: Python itself does not support function overloading. We used the `inspect` module to support overloading in hidet. This is needed because some pytorch function/methods have multiple signatures. 
The implementation can be found at [here](https://github.com/hidet-org/hidet/blob/8d755f73de0ebc109eb7dbccb825a9471a0e8811/python/hidet/graph/frontend/torch/interpreter.py#L29-L55) and [here](https://github.com/hidet-org/hidet/blob/8d755f73de0ebc109eb7dbccb825a9471a0e8811/python/hidet/graph/frontend/torch/interpreter.py#L118-L135).

xinli-git: nice!

hjjq: I see. But does `torch.Tensor.sum` and `torch.sum` have the same signature? https://pytorch.org/docs/stable/generated/torch.sum.html#torch.sum
It doesn't seem that either of them has the `out` argument.

xinli-git: > Also it doesn't seem that either of them has the out argument.
right, let me fix this

I think the overload is for sum(x, *, dtype) and sum(x, dims, keepdim, ...)?

yaoyaoding: I usually jump to the signatures in the python code to check the variants of the torch functions:
image
and its interesting that the code has `out` parameter but the documentation does not have.

hjjq: > I think the overload is for sum(x, *, dtype) and sum(x, dims, keepdim, ...)?

I see.
Also, keepdim in the first case (L989) should default to False?
Lastly, `torch.Tensor.sum` seems to have a slightly different signature, where `dim` has a default value (whereas torch.sum doesn't have a default, making `dim` mandatory). So in the case below, it will resolve to the first case because of missing `dim`, and possibly produce wrong results?
```
a = torch.randn(...)
b = a.sum(keepdim=True)
```

hjjq: > I usually jump to the signatures in the python code to check the variants of the torch functions: image and its interesting that the code has `out` parameter but the documentation does not have.

image
Interestingly, my pytorch code doesn't have `out`. Maybe we have different version/build of torch.

xinli-git: It is weird, I see two places where Torch generates
```python
@overload
def xxx(args, )
```
One under `./_C/__init__.pyi` (does not have "out"), another under `./_C/_VariableFunctions.pyi` (has "out"). However, both are just generated signatures that don't really represent the actual implementation. I think they are there to make the IDEs work.

The actual implementation should be at aten/src/ATen/native/native_functions.yaml, which has "out".

Even if the actual op does not support "out", having an optional out argument should not break the `inspect.Signature.bind` function. so we should still be fine, and it would be better to include "out" here

PR #346, opened 2023-08-09
[Graph][IR] Preliminary Hidet IR Grammar

Aalanli:
Looking for suggestions.
- Currently I try to encode the minimum set of things, so I removed some things that would cause ambiguity, such as SeqStmt.
- Things that are not represented via the grammar rules can be represented as an Attribute, such as layouts.
- Not whitespace significant
- For more see Minutiae of grammar.md


python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+( t | $\epsilon$ ) : Optional
+
+#### Minutiae
+- `@` to prevent ambiguity with Call
+- Every statement has an optional `#` IDENT, which encodes an optional attribute dict, the same attribute dict is defined in the top level module space
+- We encode things that are not part of the grammar as an attribute, for example, layouts. This keeps everything relatively extensible
+
+## Grammar
+Value := STRING | INT | FLOAT | Dict | List | Tuple
+Pair := STRING `:` Value
```
yaoyaoding, line 23: I prefer 'DictItem' here.

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+Attribute := AttributeName `=` Dict `;`
+
+DataType := `i1` | `i8` | `i16` | `i32` | `i64` | `u8` | `u16` | `u32` | `u64` | `f16` | `f32` | `f64` | `void` | `str`
+PtrType := `~`DataType
+TensorType := DataType`<` NAT (`,` NAT)\* (`,` AttributeName | $\epsilon$ ) `>`
+TensorPtrType := `~`TensorType
+Type := DataType | PtrType | TensorType | TensorPtrType
+
+Module := `{` ( Function | Attribute )\* `}` 
+Function := `def` IDENT `(` ( Arg`:` Type | $\epsilon$ ) (`,` Arg`:` Type)\*` )` `->` Type `{` Stmt* `}`
```
yaoyaoding, line 38:
```suggestion
Function := `def` IDENT `(` ( Arg`:` Type | $\epsilon$ ) (`,` Arg`:` Type)* `)` `->` Type `{` Stmt* `}`
```

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+ReturnStmt := `return` Expr `;`
+LetStmt := `let` IDENT `:` Type `=` Expr  ( `in` Stmt | $\epsilon$ ) `;`
+ForStmt := `for` ( IDENT | `(` IDENT (`,` IDENT)\* `)` ) `in` Mapping `{` Stmt* `}`
+WhileStmt := `while` Expr `{` Stmt* `}`
+BreakStmt := `break` `;`
+ContinueStmt := `continue` `;`
+IfStmt := `if` Expr `{` Stmt* `}` ( `else` `{` Stmt* `}` | $\epsilon$ )
+AssertStmt := `assert` Expr `;`
+BlackBoxStmt := `BlackBox` `{` STRING `}`
+AsmStmt := ( `volatile` | $\epsilon$ ) `asm` `{ in {` IDENT (`,` IDENT) `}` `out` `{` IDENT (`,` IDENT) `}` `{` STRING `}` `}`
```
yaoyaoding, line 66: Better to use similar syntax like https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html#assembler-asm-statements

yaoyaoding: Put the asm string in the front of the stmt.

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+BinaryOp := `+` | `*` | `-` | `/` | `//` | `%` | `**` | `^` | `<<` | `>>` | `||` | `&&` | `==` | `!=` | `<` | `>` | `<=` | `>=`
```
yaoyaoding, line 69: We do not have `//` in hidet ir.

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+UnaryOp := `-` | `!` 
+Expr := `(` Expr BinaryOp Expr `)`
+		| `(` UnaryOp Expr `)`
```
yaoyaoding, line 72: Better not add the () here.
Use something like
```
Expr := Expr BinaryOp Expr
  | UnaryOp Expr
  | `(` Expr `)`
```

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+		| GetItem
+		| TensorSlice
+		| Call
+		| Let
+		| Cast
+		| Constant
+		| IfThenElse
+		| Dereference
+		| Address
+		| Reference
```
yaoyaoding, line 82: Currently, do not add `Reference` as an expression. I will double check whether we need it.

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,101 @@
...
+IfThenElse := `if` Expr `then` `{` Expr `}` `else` `{` Expr `}`
+Dereference := `deref` Expr
+Address := `addr` Expr
+Reference := `ref` Expr
+
+Mapping := SpatialTaskMapping | RepeatTaskMapping | ComposedTaskMapping
+SpatialTaskMapping := `@spatial` `(` `shape` `=` `[` INT (`,` INT)\* `]` `,` `ranks` `=` `[` INT (`,` INT)\* `]` `)`
+RepeatTaskMapping := `@repeat` `(` `shape` `=` `[` INT (`,` INT)\* `]` `,` `ranks` `=` `[` INT (`,` INT)\* `]` (`,` AttributeName | $\epsilon$ ) `)`
+ComposedTaskMapping := `@compose` `(` Mapping `,` Mapping `)`
+
```
yaoyaoding, line 101: Better also add the layout specification.

yaoyaoding: And add it as an optional argument in TensorType. We can opt it if it has row major layout.

yaoyaoding (commented):
Thanks @Aalanli!

It looks good to me in general. Left some minor suggestions.

python/hidet/ir/Grammar.md
Aalanli: I think that the escape sequence wasn't working before, I fixed it.

python/hidet/ir/Grammar.md
Aalanli: Like this?

python/hidet/ir/Grammar.md
Aalanli: Unfortunately, this makes the grammar ambiguous, so it would take some non-trivial restructing to preserve order of operations and make the grammar unambiguous.

yaoyaoding: I see. 
When using antlr4, we can use the order of the derivatives to specify their priority, like [here](https://github.com/yaoyaoding/csc488/blob/a273aae43f1f774c23f833f6d7ace5edb61dce54/assignments/a2/minicc/grammars/MiniC.g4#L70-L86).

yaoyaoding: Not sure whehter the parser generator we are gonna to use support this feature.

python/hidet/ir/Grammar.md
yaoyaoding: Yeah.

python/hidet/ir/Grammar.md
Aalanli: Layouts can be handled by attributes, which are compile time data.

yaoyaoding: Task mapping is similar like layout. In hidet, we can use expression as the layout/task mapping argument like
```python
def matmul(ptr: ~float32, m: int, n: int):
   a = tensor_pointer(ptr, shape=[m, n], layout=row_major(m, n))
```
One solution is to treat the task mapping and layout construction like normal function calls. In the AST level, we do not distinguish them. In this way, we can have
```
ForMappingStmt := for indices in EXPR `.` `on` `(` EXPR `)`
```

python/hidet/ir/Grammar.md
yaoyaoding: Can I know whether the lark support that feature like antlr4?

I have see the new grammar about expression, and I feel it is not as readable as the one that use order to determine priority. And it might increase the difficulty to convert the AST to hidet ir if you define it in this way. But I will leave it to you to decide (or you can come back to update the grammar when you implement the conversion).

yaoyaoding:
Another suggestion is to move the grammer to `hidet.ir` submodule.

python/hidet/ir/Grammar.md
Aalanli: No, I don't believe that it is possible in lark.
Alright, I am working on the IR dumper right now, since by default it dumps all binary and unary expressions in brackets, I am not sure if it is necessary. Since to remove the reliance of always printing brackets I would have to inspect the parse tree ahead of time to determine the order of operations.

python/hidet/ir/Grammar.md
Aalanli: Then can we assume that only `SymbolVar` can be parametrized as a layout parameter in a function type signature? 
eg:

```python
import hidet
from hidet.ir.type import tensor_type
from hidet.lang import attrs, f16, tensor_pointer
from hidet.lang.layout import row_major
from hidet.lang.cuda import dynamic_shared_memory

asym = hidet.symbol_var('a', 'int32')
test_layout = row_major(64, asym)

smem_a_type = tensor_type(
        'float16', shape=[64, asym], layout=test_layout
    )

with hidet.script_module() as module:
    @hidet.script
    def test_fn(a: smem_a_type):
        a[0, 0] += 1
    
    @hidet.script
    def main(a: ~f16):
        attrs.cuda.grid_dim = 1
        attrs.cuda.block_dim = 32
        attrs.cuda.dynamic_smem_bytes = 64 * asym
        smem_a = tensor_pointer(
            'float16', shape=[64, asym], layout=test_layout
        )
        smem_a = dynamic_shared_memory(byte_offset=0, dtype=float16)
        test_fn(smem_a)

ir_module = module.ir_module()
print(ir_module)
tst = ir_module.build()
```

yaoyaoding: Better not, it can also be normal Var. If the user want, they can also use expressions.

python/hidet/ir/Grammar.md
yaoyaoding: In ANTLR4, we can define the grammar like
```
BINARY_EXPR := EXPR * EXPR
            |= EXPR + EXPR
```
Then, the parser will generate AST for "a+b*c" as <ADD(a, MUL(b, c))> and "a * b + c" as <ADD(MUL(a, b), c)>. 

We better specify the operator priority in the grammar, instead of analying after transformation (it that's possible). 

Also, it is not a good idea to force the text form to always have "()" like "(a+(b+((e+f)*d)))".

yaoyaoding: The python grammar implemented in lark: https://github.com/lark-parser/lark/blob/master/examples/advanced/python2.lark

yaoyaoding: If you find lark is not powerful enough, we can also consider using antlr4: https://alanhohn.com/posts/2016/antlr-python/
(if we use antlr4, we can consider put both the grammar and the generated python source code in our git repo).

python/hidet/ir/Grammar.md
```diff
@@ -0,0 +1,118 @@
...
+	| BOOL
+	| NUMBER
+	| STRING
+	| IDENT
+
+
+Mapping := SpatialTaskMapping | RepeatTaskMapping | ComposedTaskMapping
+SpatialTaskMapping := `@spatial` `(` `shape` `=` `[` INT (`,` INT)\* `]` `,` `ranks` `=` `[` INT (`,` INT)\* `]` `)`
+RepeatTaskMapping := `@repeat` `(` `shape` `=` `[` INT (`,` INT)\* `]` `,` `ranks` `=` `[` INT (`,` INT)\* `]` (`,` AttributeName | $\epsilon$ ) `)`
+ComposedTaskMapping := `@compose` `(` Mapping `,` Mapping `)`
```
yaoyaoding, line 118: The shape of task mapping is not necessary constant integer. They can be any integer expressions.

yaoyaoding:
Close as #354 has been merged.

PR #347, opened 2023-08-11
[Torch][Graph][Operator] Add and fix various items for torchvision model support

hjjq:
1. Enhance support for `__setitem__` and` __getitem__` of Tensor; Add SetStridedSlice Op, Roll Op.
2. Add/Update torch mapping for adaptive_avg_pool3d, eq, pad, roll, matmul, new_zeros, batch_norm, MultiHeadAttention.
3. Update torch Linear mapping to optionally accept transposed weights.
4. Fix a bug where a empty graph will output a zero tensor instead of the input/weight.


yaoyaoding (approved):
It looks good to me, good job!

PR #348, opened 2023-08-15
[Torch][Operator] More torchvision model support

hjjq:
This is a continuation of #347.

1. Add LP normalization task (ToDo: schedule template)
2. Add torch mappings for normalize, clone, zero_, exp, chunk
3. Add ceil_mode=True support for pool2d
4. Fix dtype issue in resize
5. Fix other bugs in pad, conv2d_pattern


yaoyaoding (approved):
Thanks @hjjq, looks good to me!

PR #351, opened 2023-08-21
[Driver] Make compilation more compatible with multi-processing #350

xinli-git:
This change adds a filelock to task compilation so that workflows such as distributed inference only builds the task once and avoids any potential data (file) races. 

Currently, only task building is included in the filelock because in general, compiled graphs will be different and compiled modules is already protected by task building.


xinli-git:
Hi @soodoshll maybe a quick review ?

PR #354, opened 2023-08-24
[IR][Parser] Hidet IR grammar, parser and ir reconstructor

Aalanli:
Currently, all IR roundtrips correctly in the test suite, for all passes. The text is identical except for the attention ir module, which has trivial changes (eg. (0 == wj) -> (wj == 0), or (!false) -> true).

Also fixed inline let stmt pass, so that types of tensors do not change midway. 

We need to add a dependency on lark, as the parser it generates is not correct (does not have the same behavior as the original).


Aalanli:
@yaoyaoding, is it possible to add a dependency on lark?

yaoyaoding:
Yeah, we can add the dependency.

python/hidet/ir/tools/ir_dumper.py
```diff
@@ -0,0 +1,1630 @@
...
+    new_lines = new.split('\n')
+    for oline, nline in zip(old_lines, new_lines):
+        if oline != nline:
+            print(green(oline))
+            print(red(nline))
+        else:
+            print(oline)
+
+
+GRAMMAR = r"""
```
yaoyaoding, line 1480: Do we need to keep the grammar in both `hidet.lark` and in the source code?

It would be easier to keep the consistency of the grammar when it is only stored at one place. Maybe we can read the file and store it to `GRAMMAR` when we want to parse a dumped ir.

tests/ir/parser/test_parser.py
```diff
@@ -0,0 +1,131 @@
...
+    h = symbol_var('h')
+    q = tensor_input('q', 'float16', [1, h, s, 64])
+    k = tensor_input('k', 'float16', [1, h, s, 64])
+    v = tensor_input('v', 'float16', [1, h, s, 64])
+    task = AttnTask('attn', q, k, v, False)
+    mod = task.implement_cuda('.')
+    return mod[0]
+
+
+def generate_ir_modules():
```
yaoyaoding, line 85: The tests are very comprehensive :)

yaoyaoding (approved):
Roughly go through the parser and dumper, and they look good to me.

Thanks for the efforts @Aalanli !

tests/ir/parser/test_parser.py
Aalanli: thanks

python/hidet/ir/tools/ir_dumper.py
Aalanli: It seems that reading files from within packages involves some fiddly work with paths. I found a solution that uses importlib_resources library, but only works for python >= 3.7, so I guess just including the grammar as a string is the simplest way.

yaoyaoding: Yeah, let's go with adding the grammar directly in the source code.

yaoyaoding:
Thanks @Aalanli !

PR #355, opened 2023-08-24
[Quantization] fix quantization pass bug

Aalanli:
For llama 7b, decoding with no prefill

fp16 - 128 tokens
org_t: 3.048099994659424s
avg_t: 0.0044784750789403915s

int8 - 128 tokens
org_t: 2.2229115962982178s
avg_t: 0.0038476772606372833s

### DIFF
--- a/python/hidet/graph/transforms/__init__.py
+++ b/python/hidet/graph/transforms/__init__.py
@@ -49,16 +49,16 @@ def optimize(graph: FlowGraph) -> FlowGraph:
     ret: FlowGraph
         The optimized flow graph.
     """
+    ctx = PassContext.current()
     passes = [
         subgraph_rewrite_pass(),
         automatic_mix_precision_pass(),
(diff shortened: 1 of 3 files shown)


python/hidet/graph/transforms/__init__.py
```diff
@@ -49,16 +49,16 @@ def optimize(graph: FlowGraph) -> FlowGraph:
     ret: FlowGraph
         The optimized flow graph.
     """
+    ctx = PassContext.current()
     passes = [
         subgraph_rewrite_pass(),
         automatic_mix_precision_pass(),
-        subgraph_rewrite_pass(),
+        subgraph_rewrite_pass(ctx.configs['quantize_patterns']),
```
yaoyaoding, line 56: Consider to write a new pass named quantize which can use subgraph rewrite to implement the functionality, and make the passes list clean.

yaoyaoding (approved):

PR #357, opened 2023-09-14
CPU AVX implementation for Softmax, Norm

fishingguy456:
Working but inefficient batch matmul. Takes path of matmul_f32_x86 instead of cpu autoscheduler.


python/hidet/graph/ops/matmul/resolve.py
```diff
@@ -96,36 +97,45 @@ class MatmulResolveRule(ResolveRule):
...
     """
 
-    def run_batch_matmul(self, a: Tensor, b: Tensor) -> Tensor:
-        parallel_k = self.get_config('parallel_k', default='default')  # 'default', 'search', 2, 4, ...
-        mma = self.get_config('mma', default='simt')  # 'simt', 'mma'
-
-        if any(not isinstance(v, int) for v in a.shape + b.shape):
-            nparts = 1
+    def run_batch_matmul(self, a: Tensor, b: Tensor, is_cpu: bool) -> Tensor:
+        if is_cpu:
```
yaoyaoding, line 101: We can directly check the device of tensor, without the need to pass `is_cpu` as parameter.
```
is_cpu = a.device.is_cpu()
```

python/hidet/graph/ops/normalize/norm.py
```diff
@@ -104,12 +105,6 @@ def norm_compute(*indices):
             attributes={'dims': dims, 'accumulate_dtype': accumulate_dtype, 'epsilon': epsilon},
         )
 
-    def allow_prologue(self) -> bool:
-        return False
-
-    def allow_epilogue(self) -> bool:
-        return True
-
```
yaoyaoding, line 112: If the cpu and cuda task have different behavior, it is better to create a subclass of the task and override the subclass:
```
class CPUNormalizeTask(NormalizeTask):
    ...
```

yaoyaoding: And implement the resolve rule to convert the normalize operator to corresponding cpu_normalize operator.

python/hidet/graph/ops/normalize/norm.py
```diff
@@ -352,6 +347,105 @@ def norm_kernel(x: dtype[x.shape], y: dtype[y.shape]):
...
+                            for j in range(8):
+                                tail_idx = spatial(*tail).map(i * 8 + j)
+                                out[head_idx][tail_idx] = temp_out[j]
+                    for i in range(tail_size % 8):
+                        tail_idx = spatial(*tail).map(tail_size - tail_size % 8 + i)
+                        out[head_idx][tail_idx] = \
+                            (x[head_idx][tail_idx] - mean) * prim.rsqrt(var + self.attrs['epsilon'])
+
+        norm_cpu_kernel.kind = "cpu_kernel"
+        avx_f32x8_find_sum.kind = "cpu_internal"
```
yaoyaoding, line 444: Better to avoid setting the function attributes outside its definition. Instead, use
```python
from hidet.lang import attrs

@hidet.script
def norm_cpu_kernel(...):
    attrs.func_kind = "cpu_kernel"
    ...
```

python/hidet/graph/ops/softmax.py
```diff
@@ -16,6 +16,9 @@
 from hidet.ir.builders import StmtBuilder
 from hidet.ir.primitives import active_mask, shfl_down_sync, shfl_sync
 from .utils import Task, TensorNode, compute, reduce
+from typing import List, Union
```
yaoyaoding, line 19: Move to the top.

Remember to run format & lint, see https://docs.hidet.org/stable/developer-guides/contributing.html#contributing

python/hidet/graph/ops/softmax.py
```diff
@@ -153,3 +156,143 @@ def softmax_kernel(xs: xdtype[shape], ys: xdtype[shape]):
         ir_module = module.ir_module()
 
         return ir_module
+
+    def implement_cpu(self, working_dir: str) -> Union[IRModule, List[IRModule]]:
+        # if not all(is_constant(dim) for dim in self.inputs[0].shape)\
```
yaoyaoding, line 161:
```suggestion
```

python/hidet/graph/ops/softmax.py
```diff
@@ -153,3 +156,143 @@ def softmax_kernel(xs: xdtype[shape], ys: xdtype[shape]):
...
+        if self.inputs[0].type.dtype != float32:
+            return NotImplemented  # use auto-scheduler
+        return tune.extract_ir_modules(self.schedule_softmax_cpu)
+
+    def allow_epilogue(self) -> bool:
+        return False
+
+    def allow_prologue(self) -> bool:
+        return False
```
yaoyaoding, line 170: Create a CPU version of the operator because cuda version allows prologue & epilogue.

python/hidet/graph/ops/softmax.py
```diff
@@ -153,3 +156,143 @@ def softmax_kernel(xs: xdtype[shape], ys: xdtype[shape]):
...
+                                last_idx = spatial(*tail).map(tail_size - tail_size % 8 + j)
+                                out[head_idx][p][last_idx] = prim.exp(x[head_idx][p][last_idx] - max_arr[j])
+                                sum_exp_arr[j] += out[head_idx][p][last_idx]
+                        for p in range(axis_size):
+                            for j in range(tail_size % 8):
+                                last_idx = spatial(*tail).map(tail_size - tail_size % 8 + j)
+                                out[head_idx][p][last_idx] = out[head_idx][p][last_idx] / sum_exp_arr[j]
+
+            softmax_cpu_kernel.kind = "cpu_kernel"
+            apply_exponent.kind = "cpu_internal"
```
yaoyaoding, line 295: ditto

python/hidet/ir/expr.py
```diff
@@ -439,6 +439,10 @@ def __init__(self, func_var, args):
         self.func_var: Var = func_var
         self.args: Tuple[Expr, ...] = args
 
+        if not (isinstance(func_var, Var) and isinstance(args, tuple)):
+            print(func_var, args)
+            print(type(args[0]))
+            print(type(func_var), type(args))
```
yaoyaoding, line 445:
```suggestion
```

python/hidet/ir/primitives/cpu/avx.py
```diff
@@ -43,6 +51,47 @@ def register_primitive_functions():
     for name, codegen_name, func_type in functions:
         register_primitive_function(name=name, func_or_type=func_type, codegen_name=codegen_name)
 
+    from hidet.lang import script, attrs
+    from hidet.ir.dtypes import f32x8, f32
+    from hidet.ir.func import Function
+
+    @script
+    def avx_x86_f32x8_find_sum(x: f32x8) -> f32:
```
yaoyaoding, line 59: Is there any convention to use "find" in the function name?

If not, I would prefer to name directly as "avx_x86_f32x8_sum" and "avx_x86_f32x8_max".

yaoyaoding (changes requested):
Thanks @fishingguy456 for your first PR to hidet!

I left some comments. In general, 
1. do not forget to run tests, lint and format before submit the PR.
2. for the new operators, we should add some tests for them. See the examples in `tests/operators/...`. 
3. our current design allow one task to have cpu and cuda implementation, and they share the same property of whether allow prologue, epilogue. When we want to change the allow properties, it is better to create a new task override the original one, so that it does not interfere with our original operator. In the future, we might want to add a device parameter to these functions (like `allow_prologue(self, device) -> bool`) so that we do not need creating a new class. But for now, let's create a new class and add resolve rule.

yaoyaoding:
Thanks @fishingguy456 ! Could you also add a test for softmax? 

Hi @BolinSNLHM, could you have a look of this PR? I did not check the kernel implementation details.

python/hidet/ir/primitives/cpu/avx.py
```diff
@@ -69,6 +87,51 @@ def register_primitive_functions():
     for name, codegen_name, func_type in functions:
         register_primitive_function(name=name, func_or_type=func_type, codegen_name=codegen_name)
 
+    from hidet.lang import script, attrs
+    from hidet.ir.dtypes import f32x8, f32
+    from hidet.ir.func import Function
+
+    @script
+    def avx_x86_f32x8_sum(x: f32x8) -> f32:
```
BolinSNLHM, line 95: Would it be possible to only declare the primitives (e.g., `avx_x86_f32x8_extract_half`) in this file, and then define functions like `avx_x86_f32x8_sum` as helper functions in Hidet Script in a separate file where it would be needed? The code should work as it is, but it looks a bit odd to have `hidet.script`decorator and multiple calls to `call_primitive_func` here...

python/hidet/ir/primitives/cpu/avx.py
```diff
@@ -69,6 +87,51 @@ def register_primitive_functions():
...
+    def avx_x86_f32x8_scalar_max(x: f32x8) -> f32:
+        attrs.func_kind = "cpu_internal"
+        attrs.func_name = "avx_x86_float32x8_scalar_max"
+        y = call_primitive_func('avx_x86_float32x8_permute_2f128', [x, x, 1])
+        m1 = call_primitive_func('avx_x86_float32x8_max', [x, y])
+        m2 = call_primitive_func('avx_x86_float32x8_permute', [m1, 0b01001110])
+        m3 = call_primitive_func('avx_x86_float32x8_max', [m1, m2])
+        m4 = call_primitive_func('avx_x86_float32x8_permute', [m3, 0b10110001])
+        m = call_primitive_func('avx_x86_float32x8_max', [m3, m4])
+        return call_primitive_func('avx_x86_float32x8_extract_last', [m])
```
BolinSNLHM, line 122: Would it be possible to only declare the primitives (e.g., `avx_x86_f32x8_extract_half`) in this file, and then define functions like `avx_x86_f32x8_sum` as helper functions in Hidet Script in a separate file where it would be needed? The code should work as it is, but it looks a bit odd to have hidet.script decorator and multiple calls to call_primitive_func here...

python/hidet/graph/ops/matmul/matmul_f32_x86.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, a: TensorNode, b: TensorNode):
         )
 
     def allow_epilogue(self) -> bool:
-        return True
+        return False
```
BolinSNLHM, line 76: Why should we change this to False? 🤔

python/hidet/ir/primitives/cpu/avx.py
yaoyaoding, line 122: I recommand to move these user-defined functions over avx (not the ones provided by underlying vector library) to another file called `avx_helpers.py`.

For functions like `avx_x86_float32x4_extract_last`, we also need to define a wrapper function like
```python
def avx_x86_float32x4_extract_last(x: Expr) -> Call:
    return call_primitive_func('avx_x86_float32x4_extract_last', [x])
```

yaoyaoding: In the new file, we directly use `avx_x86_float32x4_extract_last(...)` in the hidet script, instead of calling `call_primitive_func`.

yaoyaoding (approved):

yaoyaoding:
Thanks @fishingguy456 !

PR #363, opened 2023-10-10
[Graph][Benchmark] Update benchmark function

Aalanli:
The old benchmarking function did not clear the l2 cache, so repeated runs are biased.
This is especially prevalent in tuning for parallel-k parts, which always selects k_parts=1 due to l2 cache hits, even when it is not the fastest implementation.


yaoyaoding:
Hi Allan,

The PR looks good to me. But before merging, it is better to have some demos on the improvement of the accuracy on the selection of parallel k when we clear the L2 cache.

Aalanli:
After some further investigation, it appears that the clearing of the l2 cache is not the greatest contributor, but the usage of `torch.cuda.Event`, which I assume to be more accurate than `time.time()`.
Here is my benchmarking script for reference: https://gist.github.com/Aalanli/b81d1a751a78ea72b491d872aa993f9e

image
image

Aalanli:
I just removed the torch dependencies.

python/hidet/cuda/device.py
```diff
@@ -172,3 +172,48 @@ def profiler_stop():
...
+        The end event.
+
+    Returns
+    -------
+    elapsed: float
+        The elapsed time in milliseconds.
+    """
+    (err, t) = cudart.cudaEventElapsedTime(start, end)
+    assert err == 0, err
+    return t
```
yaoyaoding, line 219: Can we use the `Event` class in https://github.com/hidet-org/hidet/blob/main/python/hidet/cuda/event.py ?

yaoyaoding (approved):

PR #365, opened 2023-10-11
[Compile Server] Update deps for compilation server

xinli-git:
There is a dependency lacking in the requirements.txt of the compilation server, causing it to fail at import. Adding it back here.


yaoyaoding:
Thanks @xinli-git !

PR #369, opened 2023-10-25
[Dynamo] Refactoring code for Hidet remote compilation

destefy:
Hidet remote compilation needs to access the FlowGraph in order to hash it, so it has been moved into a separate function (get_flow_graph)
Remote compilation also needs to access the CompilationGraph since it will send that from the server to the client, so it as been moved into a separate function (get_compiled_graph)
Since the remote compilation client receives the CompilationGraph, the executor and wrapper function that uses this cgraph are moved to a separate function (get_wrapper)

### DIFF
--- a/python/hidet/graph/frontend/torch/dynamo_backends.py
+++ b/python/hidet/graph/frontend/torch/dynamo_backends.py
@@ -10,28 +10,65 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # pylint: disable=no-name-in-module
-from typing import List, Callable, Sequence, Union
+from typing import List, Sequence, Union
 import logging
 import torch
 import hidet.option
+from hidet import Tensor
+from hidet.ir import dtypes
 from hidet.ir.type import DataType
 from hidet.ir.expr import SymbolVar
+from hidet.runtime import CompiledGraph
 from hidet.graph.flow_graph import FlowGraph
 from hidet.graph.transforms import PassContext, optimize
-from hidet.runtime import CompiledGraph
 from hidet.cuda.graph import CudaGraphCreationError
-from hidet.ir import dtypes
-from .utils import serialize_output, deserialize_output, resolve_save_dir_multigraph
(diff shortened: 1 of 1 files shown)


python/hidet/graph/frontend/torch/dynamo_backends.py
```diff
@@ -10,28 +10,65 @@
...
+            inputs.append(symbolic_input)
+        elif isinstance(example_input, (int, bool, float)):
+            inputs.append(example_input)
+        elif isinstance(example_input, torch.SymInt):
+            from torch.fx.experimental.symbolic_shapes import SymNode
+
+            node: SymNode = example_input.node
+            try:
+                inputs.append(node.pytype(example_input))
+            except RuntimeError:
```
xinli-git, line 48: should we be concerned here? 
RuntimeError seems to be quite generic so it might not just be the error that you think you can handle. I suggest we log a warning down below if it does not happen too often. 

If possible, I prefer that you check the pre-condition that would cause the runtime error and instead of try-except, use 
```
if not (the condition that would trigger runtime error): 
  inputs.append(...)
else:
  # is a symbolic scalar input
```

xinli-git (commented):
Can you follow the PR title format and add a description?

python/hidet/graph/frontend/torch/dynamo_backends.py
xinli-git: I see that this part is ported from the original code so you might not know why. Please ignore it for now

xinli-git (approved):
Thanks for the PR!

python/hidet/graph/frontend/torch/dynamo_backends.py
yaoyaoding: Hi @xinli-git, FYI, we (Allan and I) have tried that, but find out it is hard to find the `the condition that would trigger runtime error`.

yaoyaoding (approved):
It looks good to me! Thanks @destefy.

Issue #370, opened 2023-10-28
[NotImplementedError] operator.lshift


HanGuo97:
**Describe the Problem**
BackendCompilerFailed: hidet_backend raised NotImplementedError: The following modules/functions are not supported by hidet yet:
  operator.lshift

yaoyaoding:
Thanks @HanGuo97 for reporting this. 

Hi @Aalanli, could you help add this operator when you have a chance?

Aalanli:
Fixed in https://github.com/hidet-org/hidet/pull/371

HanGuo97:
Thanks for the quick fix!

Does that mean I could simply build from source, and this would work via `torch.compile` interface?

Aalanli:
I am not sure about the particulars of your model, but this script worked for me (when built from source):
```python
```

HanGuo97:
Hi @Aalanli, 
Thank you so much for your help! The patch you provided worked perfectly for me. 

I wanted to follow up with a more detailed example. This particular issue caused several errors for me. Some of these were fixable using the patch you provided, such as `pow`, `bitwise_and`, and `torch.Tensor.max`. 

However, there is one error that I’m having difficulty fixing without additional knowledge of the codebase. Here’s the error message: 

> ValueError: Unknown data type: uint8x4, candidates...

I’ve attached a simplified code that shows this error, and possibly a few others. I would appreciate any help you can offer in resolving this issue. Thank you in advance for your time!


<details>
<summary>Code</summary>

```python
import math
import torch
import jaxtyping
from typing import Tuple

DEFAULT_CONTAINER_NUM_BITS = 8
FloatTensorType = jaxtyping.Float[torch.Tensor, "..."]
UInt8TensorType = jaxtyping.UInt8[torch.Tensor, "..."]
BinaryTensorType = jaxtyping.Bool[torch.Tensor, "..."]
PackedBinaryTensorType = jaxtyping.UInt8[torch.Tensor, "..."]


def from_binary(tensor: BinaryTensorType, num_bits: int) -> UInt8TensorType:
    if tensor.dtype != torch.bool:
        raise TypeError
    if tensor.shape[-1] != num_bits:
        raise ValueError
    if num_bits > 8:
        raise NotImplementedError
    mask = torch.tensor([2], dtype=torch.float32, device=tensor.device) ** torch.arange(
        num_bits - 1, -1, -1,
        dtype=torch.float32,
        device=tensor.device)
[...]

PR #372, opened 2023-10-30
[Graph][Operator] Fix reduce bug, add uint8x4

Aalanli:
1. Fix reduce bug in vector type resolution
2. Add support for uint8x4 vector type


Aalanli:
Hi @yaoyaoding , any comments before I merge this? This is for issue https://github.com/hidet-org/hidet/issues/370.

yaoyaoding (approved):
Thanks @Aalanli !

It looks good to me in general. A minor suggestion is to add some tests for the new data type and operator.

PR #377, opened 2023-11-14
[CompiledGraph] Add option to store dispatch table option

destefy:
This option will include the dispatch_table file as part of save_compiled_graph

This way, it will be extracted into the hidet/graphs cache when load_compiled_graph is called so the client can avoid this fine-tuning

### DIFF
--- a/python/hidet/graph/frontend/torch/dynamo_backends.py
+++ b/python/hidet/graph/frontend/torch/dynamo_backends.py
@@ -74,7 +74,6 @@ def get_compiled_graph(flow_graph: FlowGraph):
     parallel_k = dynamo_config['parallel_k']
     tensor_core = dynamo_config['use_tensor_core']
     save_dir = dynamo_config['dump_graph_ir']
-
     with PassContext() as ctx:
         if use_fp16:
             ctx.set_precision('float16')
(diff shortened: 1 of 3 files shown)


python/hidet/graph/frontend/torch/dynamo_backends.py
```diff
@@ -98,6 +97,17 @@ def get_compiled_graph(flow_graph: FlowGraph):
...
+    torch_inputs: List[torch.Tensor] = []
+    for x in inputs:
+        if not x.is_contiguous():
+            # warnings.warn_once('Hidet received a non-contiguous torch input tensor, converting it to contiguous')
+            x = x.contiguous()
+        torch_inputs.append(x)
+    hidet_inputs: List[hidet.Tensor] = [hidet.from_torch(tensor) for tensor in torch_inputs]
+    return hidet_inputs
+
+
```
yaoyaoding, line 110: Let's only keep the necessary change to the purpose of this PR.

python/hidet/option.py
```diff
@@ -263,6 +263,12 @@ def register_hidet_options():
         default_value=True,
         description='Whether to enable imperative execution when op arguments allows',
     )
+    register_option(
+        name='store_dispatch_table',
+        type_hint='bool',
+        default_value=False,
+        description="Whether to save and load the CompiledGraph's dispatch table",
+    )
```
yaoyaoding, line 271: I prefer to put the flag as a parameter to `save_compiled_graph` instead taking it as an option. Otherwise, our options will explode soon. Usually, only use `option` to place the option that will be accessed in multiple places across the whole project, or the ones that are very important and need to be exposed to users.

python/hidet/runtime/compiled_graph.py
```diff
@@ -416,6 +416,10 @@ def _save_under(dir_path: str, dir_in_zip: str, exclude: Optional[List[str]] = N
             ge_bytes = json.dumps(asdict(model.graph_execution), indent=4).encode('utf-8')
             f.write(ge_bytes)
 
+        # save dispatch table file
+        if hidet.option.get_store_dispatch_table() and os.path.exists(model.dispatch_table_path):
+            zf.write(model.dispatch_table_path, arcname="dispatch_table.txt")
```
yaoyaoding, line 421: Let's keep the way of adding dispatch table to zip file consistent with writing other files like `graph_string.txt` and `graph_execution.json`.

yaoyaoding (changes requested):
Thanks @destefy, I left some suggestions.

yaoyaoding (approved):

yaoyaoding:
Thanks @destefy !

PR #384, opened 2023-12-02
[Graph] Fix CompiledGraph aliasing bug

Aalanli:
Some graphs produce output tensors that alias. In the original case, the outputs of the aliasing tensors would be undefined (the result of empty). This PR fixes that by aliasing the outputs as well.

For example this script originally outputs garbage in the second tensor
```
import hidet

a = hidet.symbol([1, 2], device='cuda')
b = a.mean(1)
g = hidet.trace_from([b, b], a)

cg = g.build()
print(cg(hidet.ones([1, 2], device='cuda'))[1])
```


python/hidet/runtime/compiled_graph.py
```diff
@@ -208,6 +208,10 @@ def _create_outputs(self, inputs):
 
         outputs = []
         for output_index, (exec_idx, sig) in enumerate(zip(self.graph_execution.outputs_index, self.meta.outputs)):
+            if sig.alias is not None:
+                outputs.append(outputs[sig.alias])
+                continue
+
```
yaoyaoding, line 214: We can do the checking alternatively like
```
            if exec_idx in self.graph_execution.inputs_index:
                outputs.append(inputs[self.graph_execution.inputs_index.index(exec_idx)])
            elif exec_idx in self.graph_execution.weights_index:
                outputs.append(self.weights[self.graph_execution.weights_index.index(exec_idx)])
            elif exec_idx != self.graph_execution.outputs_index.index(exec_idx): # <--------
                outputs.append(outputs[self.graph_execution.outputs_index.index(exec_idx)])
            else:
```
so we do not need to modify the graph building process or an alias attribute.

How do you think @Aalanli ?

yaoyaoding (commented):
Hi @Aalanli,

Thanks for the fix! There is an alternative way to fix this that you can consider. Besides, let's add the tests provided by the @Azyka to our test bed.

commutative_law.py
```diff
@@ -0,0 +1,94 @@
+# %%
```
yaoyaoding, line 1: Remove this file.

python/hidet/runtime/compiled_graph.py
Aalanli: The suggested change doesn't work, but I see how to fix it. 
Yeah, I think this way is better as well.

yaoyaoding:
Thanks @Aalanli !

yaoyaoding:
Thanks @Aalanli !

PR #392, opened 2023-12-12
[Fix][Graph] Write compiled graph to tempfile first

destefy:
When a process is running save_compiled_graph(path) and writing the CompiledGraph zipfile to the disk, another process may think that the CompiledGraph file can be read since os.path.isfile(path) evaluates to true.

This change writes the file to a temporary file first to avoid this race condition.


yaoyaoding:
Thanks @destefy ! It looks good to me.

PR #393, opened 2023-12-14
[Option] Add option to set class Var id attribute to 0 by default

destefy:
By default Var instances will have an id of 0.

This is to ensure that each entire FlowGraph is entirely deterministic given the same interpreter+input.

### DIFF
--- a/python/hidet/ir/expr.py
+++ b/python/hidet/ir/expr.py
@@ -19,6 +19,7 @@ import numpy as np
 from .node import Node
 from .type import BaseType, TensorType, DataType, TensorPointerType, PointerType, FuncType, StringType, ArrayType
 from .type import tensor_pointer_type, string_type, tensor_type, data_type
+import hidet.option
 
 PyScalar = Union[bool, int, float, complex, str]
 
@@ -561,7 +562,7 @@ class Var(Expr):
         self.hint: Optional[str] = hint
         self.name: Optional[str] = name
         self.type: Union[BaseType, TensorType, TensorPointerType, FuncType] = type
-        self.id: int = self.new_id()
+        self.id: int = self.new_id() if hidet.option.get_option('debug_show_var_id') else 0
 
     @staticmethod
     def new_id():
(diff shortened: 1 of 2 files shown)


python/hidet/ir/expr.py
```diff
@@ -561,7 +562,7 @@ def __init__(self, hint: Optional[str], type: BaseType, name: Optional[str] = No
         self.hint: Optional[str] = hint
         self.name: Optional[str] = name
         self.type: Union[BaseType, TensorType, TensorPointerType, FuncType] = type
-        self.id: int = self.new_id()
+        self.id: int = self.new_id() if hidet.option.get_option('debug_show_var_id') else 0
```
yaoyaoding, line 565: It is better to put the logic into the `new_id` function to make the `__init__` looks more clean.

python/hidet/option.py
```diff
@@ -701,12 +709,28 @@ def debug_cache_tuning(enabled: bool = True):
...
+    OptionContext.current().set_option('debug_enable_var_id', enable)
+
+
 def debug_show_var_id(enable: bool = True):
     """
     Whether to show the var id in the IR.
 
     When this option is enabled, the IR will show the var id with the format `var@id`, like `x@1` and `d_1@1732`.
     Variable (i.e., hidet.ir.Var) a and b is the same var if and only if `a is b` evaluates to True in Python).
+    Hint: all variable ids will always be 0 unless the debug_enable_var_id option is set to True.',
```
yaoyaoding, line 733: Maybe check whether the current `debug_enable_var_id` is True or not when we turn on `debug_show_var_id`. If `debug_enable_var_id` is off, we can use
```python
warnings.warn("Please use `hidet.option.debug_enable_var_id()` to enable the id first")
```
to ask the user to turn on the id tracking.
Alternatively, we can silently turn it on.

yaoyaoding (commented):
Thanks @destefy!
Left two minor suggestions.

yaoyaoding:
Thanks @destefy!

PR #394, opened 2023-12-18
[CI] Add CI workflow and scripts

hjjq:
Also add support for cpu arch option to allow Compile Server to target a specific CPU arch.
Ubuntu ver. for Compile Server is bumped for compatibility with newer CPU gens.
Numpy ver. is bumped for compatibility.


yaoyaoding:
Thanks @hjjq! 

This PR looks good to me.

PR #399, opened 2023-12-28
[Operator] Support advanced options for pooling operators

yaoyaoding:
1. Support count_include_pad and divisor_override in average pool.
2. Refactor avg and max pool operators computation definition

PR #402, opened 2023-12-30
[Fixbug] Fix a performance bug in auto-scheduler

yaoyaoding:
Introduced in #399


yaoyaoding:
```
$hidet-ci launch
```

yaoyaoding:
$hidet-ci launch

PR #422, opened 2024-01-23
[Operator] Add cublas to matmul tune space

hjjq:
Also fix bug in call graph generation.


hjjq:
$hidet-ci launch

hjjq:
$hidet-ci launch

PR #423, opened 2024-01-31
[cuBLAS] Add cublas_gemm_batched and use cublasSetStream to set stream to the current stream in all cublas API calls

yudi0201:


src/hidet/runtime/cuda/cuda.cpp
```diff
@@ -79,3 +88,18 @@ DLL void hidet_cuda_set_device(int device) {
     lazy_load_cuda_runtime();
     CHECK_CUDA(cudaSetDevice(device));
 }
+
+DLL void hidet_cuda_malloc(void **devPtr, size_t size) {
```
yaoyaoding, line 92: Consider directly return the allocated memory address like
```c++
DLL void* hidet_cuda_malloc(size_t size) {
    ...
}
```

yaoyaoding: Like `hidet_cuda_get_device(...)`.

yaoyaoding (approved):
Thanks @yudi0201 !

It looks good to me.

src/hidet/runtime/cuda/cublas.cpp
```diff
@@ -319,4 +345,70 @@ DLL void hidet_cublas_strided_gemm(
...
+
+    set_alpha_beta(&p_alpha, &p_beta, cublasComputeType_t(compute_type), cudaDataType_t(tc));
+
+    static void **ptr_a_device, **ptr_b_device, **ptr_c_device;
+    static int cur_device_ptr_size; // Size of device memory currently allocated for each of the three a,b,c arrays.
+
+    // Allocate device memory
+    // first use synchronous versions of malloc and memcpy, later switch to async versions
+    if (cur_device_ptr_size != 0 && b > cur_device_ptr_size) {
+        hidet_cuda_free((void *)ptr_a_device);
```
yaoyaoding, line 369: Why not `hidet_cuda_free_async`?

yaoyaoding: The following logic is more readable to me, just as a reference.
```
if(b > cur_device_ptr_size) {
  if(cur_device_ptr_size > 0) {
    free the three ptrs
  }
  alloc three ptrs
}
```

src/hidet/runtime/cuda/cuda.cpp
```diff
@@ -79,3 +94,32 @@ DLL void hidet_cuda_set_device(int device) {
     lazy_load_cuda_runtime();
     CHECK_CUDA(cudaSetDevice(device));
 }
+
+DLL void* hidet_cuda_malloc(size_t size) {
+    lazy_load_cuda_runtime();
+    void* devPtr = malloc(sizeof(void*));
```
yaoyaoding, line 100:
```suggestion
    void* devPtr;
```
We do not need to allocate a memory region.

src/hidet/runtime/cuda/cuda.cpp
```diff
@@ -79,3 +94,32 @@ DLL void hidet_cuda_set_device(int device) {
...
+    CHECK_CUDA(cudaMalloc(&devPtr, size));
+    return devPtr;
+}
+
+DLL void* hidet_cuda_malloc_async(size_t size, cudaStream_t stream) {
+    lazy_load_cuda_runtime();
+    void* devPtr = malloc(sizeof(void*));
```
yaoyaoding, line 107: Same as the previous one.

yaoyaoding (commented):
Thanks @yudi0201!

src/hidet/runtime/cuda/cuda.cpp
yudi0201: Maybe I'm missing something, but wouldn't doing this result in devPtr being created on the stack, and then we'd be returning a local stack variable?

src/hidet/runtime/cuda/cublas.cpp
yudi0201: Thanks for the suggestions! I'll modify these in the next revision.

src/hidet/runtime/cuda/cuda.cpp
yaoyaoding: When we call 
```c++
cudaMallocAsync(&devPtr, ...)
```
we are passing the address of the pointer to the cuda api function, which will update the pointer value. The pointer is a stack variable and will be valid during we calling the cuda api function. We returns the "value" of the pointer instead of the "address" of the pointer to the callee of `hidet_cuda_malloc`, this is fine.

yudi0201: Got it. Thanks!

yaoyaoding:
Thanks @yudi0201 !

PR #428, opened 2024-02-20
[Fixbug] Fix graph metadata hash

KTong821:
Same graphs should have the same hash. Graph node traversal used in generating hash is non-deterministic, leading to different hashes for the same graph.

This can prevent graph runs from using the fast path, since dispatch tables cannot be found for the current (different) hash!

### DIFF
--- a/python/hidet/drivers/build_graph.py
+++ b/python/hidet/drivers/build_graph.py
@@ -145,6 +145,11 @@ def get_graph_meta_data(graph: FlowGraph, num_kernels, space: int) -> GraphMetaD
         lines.append(str(node.task))
     lines.append(str(graph))
     lines.append(str(space))
+
+    # graph nodes are not traversed in deterministic order
+    # sort to ensure same graph --> same hash
+    lines.sort()
+
     graph_hash = sha256('\n'.join(lines).encode('utf-8')).hexdigest()[:16]
 
     return GraphMetaData(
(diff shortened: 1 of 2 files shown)


yaoyaoding:
Very interesting!

Is there any example to reproduce the behavior of producing different hashes?

KTong821:
@yaoyaoding I am able to observe this on my resnet app branch, with >30 layers. Was not able to observe on smaller graphs.

python/hidet/drivers/build_graph.py
```diff
@@ -145,6 +145,11 @@ def get_graph_meta_data(graph: FlowGraph, num_kernels, space: int) -> GraphMetaD
         lines.append(str(node.task))
     lines.append(str(graph))
```
yaoyaoding, line 146: Will this line produce different string when the topological order is different?

vadiklyutiy:
@KTong821 What is the reason of Graph node traversal used in generating hash is non-deterministic?

yaoyaoding:
> using iteration over unordered data structure (set, dict), causing this behaviour.

Good catch! We might need to avoid iterating set when we construct the topological order of nodes of flow graph.
Using dict is fine because "Dictionaries preserve insertion order." from https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

@vadiklyutiy for your question, we used set and iterate it when we compute a topological order of our computation graph.

python/hidet/drivers/build_graph.py
KTong821: Yes, I think anything that relies on `graph.nodes` will cause this.

yaoyaoding:
Thanks! @KTong821

PR #435, opened 2024-03-07
[cuDNN] Add cudnn conv2d

yudi0201:


src/hidet/runtime/cuda/cudnn.cpp
```diff
@@ -0,0 +1,616 @@
...
+                CUDNN_TYPE_BACKEND_DESCRIPTOR, 1, &engcfg));
+    CHECK_CUDNN(cudnnBackendFinalize(plan));
+
+    int64_t workspaceSize;
+    CHECK_CUDNN(cudnnBackendGetAttribute(plan, CUDNN_ATTR_EXECUTION_PLAN_WORKSPACE_SIZE,
+                CUDNN_TYPE_INT64, 1, NULL, &workspaceSize));
+    
+    void *dev_ptrs[3] = {ptr_x, ptr_w, ptr_y}; // device pointers
+    int64_t uids[3] = {'x', 'w', 'y'};
+    void *workspace = hidet_cuda_malloc_async(workspaceSize, cur_stream);
```
yaoyaoding, line 588: It might be better to use the workspace shared by all hidet operators (i.e., https://github.com/hidet-org/hidet/blob/main/include/hidet/runtime/cuda/context.h#L46). 

When we run the operator in the second time, there will not be any memory allocation. Thus, it can also be used in cudaGraph.

src/hidet/runtime/cuda/cudnn.cpp
```diff
@@ -0,0 +1,616 @@
...
+    hidet_cuda_free_async(workspace, cur_stream);
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(xDesc));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(wDesc));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(yDesc));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(cDesc));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(fprop));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(op_graph));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(engine));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(engcfg));
+    CHECK_CUDNN(cudnnBackendDestroyDescriptor(plan));
```
yaoyaoding, line 613: Might be good to benchmark the performance of our implementation vs. PyTorch's conv2d performance. I am not sure whether the overhead of creating/destroying descriptors is large enough to influence the performance.

yaoyaoding (commented):
Thanks @yudi0201 ! 

Overall looks good to me. After merging this PR, we can add a primitive function to call the conv2d_cudnn in our runtime library and have an operator like `hidet.ops.conv2d_cudnn`.

vadiklyutiy:
@yaoyaoding 
A little bit different but connected question.
Are we planing to include conv2d_cudnn to search space? I mean we can search via our current space + cudnn implementation. Is it doable at all (wo huge redesign)?

yaoyaoding:
It's doable, similar to cublas gemm: https://github.com/hidet-org/hidet/commit/072a606111435b2fb46678577f291f4a4ac91d5c

vadiklyutiy:
What about adding cudnn*, cublas etc to search space?

yaoyaoding:
> What about adding cudnn*, cublas* etc to search space?

That's exactly what the commint I mentioned before does.

yaoyaoding:
Hi @c-fteixeira, the ci seems can not initialize the vm for test, could you help us to take a look? thank you!

yaoyaoding (approved):
LGTM, thanks @yudi0201 !

PR #437, opened 2024-03-09
[Fixbug] Fix for softmmax cpu causing issues

fishingguy456:
moved implement_cpu to the cpu task


vadiklyutiy:
@yaoyaoding 
Do you remember about this changes? Do we need it?

yaoyaoding:
Seems there is some problem in the non-fp32 softmax but I don't remember the exact problem. But it's okay to close this and add a PR to fix the problem by fixing the operator template.

fishingguy456:
I don't remember the issue exactly but I think it had something to do with the kernel working in isolation but not when it was included in a larger model graph because I put one of the functions in the wrong place. The change is simple so it can just be incorporated in another PR.

PR #438, opened 2024-03-13
[Dynamo] Refactor get_wrapper and pickling compiled graph

destefy:
The CentML compilation backend I am working on wants to wrap the `CompiledGraph`s forward function (the one returned by `get_wrapper`) in a `torch.fx.GraphModule`. This `GraphModule` would then be pickled and sent from a server to a client.

However, it isn't possible to pickle the lambda/local function returned by `get_wrapper`. Therefore, I am turning `get_wrapper` into a class `CompiledForwardFunction` whose `forward` function behaves like the `wrapper` returned by `get_wrapper`.

Additionally, in order to pickle `CompiledForwardFunction`, I have defined pickling and unpickling behaviour for `CompiledGraph` using `__getstate__` and `__setstate__` respectively. These just call `CompiledGraph`'s existing `save` and `load` functions.

### DIFF
--- a/python/hidet/graph/frontend/torch/dynamo_backends.py
+++ b/python/hidet/graph/frontend/torch/dynamo_backends.py
@@ -97,7 +97,6 @@ def get_compiled_graph(flow_graph: FlowGraph):
     logger.info('finish building computation graph')
     return cgraph
 
-
 def preprocess_inputs(inputs: Sequence[torch.Tensor]) -> List[hidet.Tensor]:
     torch_inputs: List[torch.Tensor] = []
     for x in inputs:
@@ -115,7 +114,7 @@ class CompiledForwardFunction(torch.nn.Module):
         self.inputs = inputs
         self.output_format = output_format
 
-    def __call__(self, *args):
+    def forward(self, *args):
         if dynamo_config['use_cuda_graph']:
             try:
                 runner = self.cgraph.cuda_graph()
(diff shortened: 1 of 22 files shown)


python/hidet/graph/frontend/torch/dynamo_backends.py
```diff
@@ -108,48 +109,59 @@ def preprocess_inputs(inputs: Sequence[torch.Tensor]) -> List[hidet.Tensor]:
...
+
+    def __call__(self, *args):
+        use_cuda_graph = dynamo_config['use_cuda_graph']
+        if use_cuda_graph:
+            try:
+                runner = self.cgraph.cuda_graph()
+            except CudaGraphCreationError:
+                runner = self.cgraph
+        else:
+            runner = self.cgraph
```
yaoyaoding: We should not create the cuda graph for every __call__ invoke.

python/hidet/graph/frontend/torch/dynamo_backends.py
```diff
@@ -108,48 +109,59 @@ def preprocess_inputs(inputs: Sequence[torch.Tensor]) -> List[hidet.Tensor]:
     return hidet_inputs
 
 
-def get_wrapper(cgraph: CompiledGraph, inputs, output_format):
-    use_cuda_graph = dynamo_config['use_cuda_graph']
-    if use_cuda_graph:
-        try:
-            runner = cgraph.cuda_graph()
-        except CudaGraphCreationError:
+class CompiledForwardFunction(torch.nn.Module):
```
yaoyaoding, line 112: I prefer the name `TorchCompiledModule`

yaoyaoding (changes requested):
Hi @destefy,

I don't like the idea of wrapping what we have returned as a `torch.fx.GraphModule`. If you want to have something that can be saved, you can use CompiledGraph. Or make the `CompiledForwardFunction` serializable.

PR #447, opened 2024-04-05
[CuBLAS] Add CuBLAS benchmarks

yudi0201:
Please refer to the commit message for benchmark results.


vadiklyutiy:
@yaoyaoding 
Do you know do we need this PR?

yaoyaoding:
We can keep this PR and integrate the benchmark to our regression test so that we know the performance comparison of cublas, hidet matmul, pytorch matmul. We can create an issue to track this and close this PR until we finish the integration.  Do we have any similar benchmark in regression test? @vadiklyutiy

vadiklyutiy:
For compare with cublas and pytorch inductor I used the following approach. I pass here 
https://github.com/hidet-org/hidet/blob/main/tests/benchmarks/run_tests.py#L56-L57 
different `backend` and `mode`. 
`backend=inductor`, `mode=eager` correspond cutlass (not cublas)
`backend=inductor`, `mode=eager` correspond inductor with triton tuning

yaoyaoding:
> For compare with cublas and pytorch inductor I used the following approach. I pass here https://github.com/hidet-org/hidet/blob/main/tests/benchmarks/run_tests.py#L56-L57 different `backend` and `mode`. `backend=inductor`, `mode=eager` correspond cutlass (not cublas) `backend=inductor`, `mode=eager` correspond inductor with triton tuning

From the commit message of this PR, we can see a more detailed comparison:
image

This comparison gives more detailed information when we use different matmul variants in cublas/cutlass, and it's on the operator level. 

By the way, the mode and backend have the same value for both cases in your comment, any typo?

vadiklyutiy:
> By the way, the mode and backend have the same value for both cases in your comment, any typo?

typo. Fixed it in original msg

vadiklyutiy:
> This comparison gives more detailed information when we use different matmul variants in cublas/cutlass, and it's on the operator level.

agree that there is more details. Maybe we should finalize this PR.

PR #448, opened 2024-04-06
[Fix] Adding header guards to header files in include/runtime/cpu

BolinSNLHM:
While trying to benchmark Hidet on the CPU, I encountered some compilation errors caused by function re-definition. It was due to multiple same `include` statements within the same `source.cc` file. 

I added the header guards to avoid the error.


vadiklyutiy:
@BolinSNLHM 
Is it actual PR?

BolinSNLHM:
> @BolinSNLHM 
> 
> Is it actual PR? 

Yes but it was before I joined the company

vadiklyutiy:
If so, maybe we can move it to CentML/hidet repo. 
From the other side CPU is not the first target right now. So, can postpone it to when it would be actual

BolinSNLHM:
> If so, maybe we can move it to CentML/hidet repo. From the other side CPU is not the first target right now. So, can postpone it to when it would be actual

Yes, that makes sense

PR #471, opened 2024-07-30
[BUILD] Build wheel with `manylinux_2_28_x86_64`

vadiklyutiy:
- build wheel with `manylinux_2_28_x86_64`
- v0.4.0 -> v0.4.1

### DIFF
--- a/.github/actions/action.yaml
+++ b/.github/actions/action.yaml
@@ -9,7 +9,7 @@ outputs:
     description: 'the path to the generated wheel'
 runs:
   using: 'docker'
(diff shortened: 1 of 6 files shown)


c-fteixeira (commented):
LGTM. 

can follow up by removing the manylinux1 folder 👍

vadiklyutiy:
> LGTM.
> 
> can follow up by removing the manylinux1 folder 👍

I agree that some cleaning is required. Not only `manylinux1` but around. 
But I think we should clean after we remake CI to use wheel for testing.

PR #488, opened 2025-01-15
[CI] chore: normalize ci runner naming

c-fteixeira:
Normalizing names and aligning runners with what is currently available.

### DIFF
--- a/.github/workflows/publish-centml-pypi.yaml
+++ b/.github/workflows/publish-centml-pypi.yaml
@@ -80,7 +80,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(needs.list-test-dirs.outputs.matrix) }}
-    runs-on: arc-runner-set
+    runs-on:
+      group: 'arc-l4'
     container:
       image: nvidia/cuda:12.4.0-devel-ubuntu20.04
     steps:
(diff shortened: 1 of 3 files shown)


vadiklyutiy:
@c-fteixeira is it exactly the same that we have in CentML/hidet?

c-fteixeira:
> @c-fteixeira is it exactly the same that we have in CentML/hidet?

Same definition on the same type of hardware, but using independent nodes.
The hardware itself never overlaps to avoid security implications that could raise.

vadiklyutiy:
> > @c-fteixeira is it exactly the same that we have in CentML/hidet?
> 
> Same definition on the same type of hardware, but using independent nodes. The hardware itself never overlaps to avoid security implications that could raise.

let me clarify. I am worry about sync repos. Both sides. So, is the code of the action byte-2-byte the same in both repos?

vadiklyutiy:
in other words all differences are outside the repo code?

c-fteixeira:
> in other words all differences are outside the repo code?

not exact, i've removed [these](https://github.com/hidet-org/hidet/pull/488/files#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191L15-L17) choices, everything else is the same. 
If you prefer, we can leave them, but right now, they would not work.

vadiklyutiy:
> > in other words all differences are outside the repo code?
> 
> not exact, i've removed [these](https://github.com/hidet-org/hidet/pull/488/files#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191L15-L17) choices, everything else is the same. If you prefer, we can leave them, but right now, they would not work.

May we leave this choices, but don't have runner for them? If somebody choose, it'll fail and this is ok. Just to be consistent between repos.

vadiklyutiy (approved):

PR #489, opened 2025-01-20
[FEATURE] Use permanent processes to handle fixed commits on the compilation server

vadiklyutiy:
Introduce permanent processes to handle fixed commits on the compilation server. 

Every high level worker might create up to 5 permanent processes for compilation of fixed version of hidet. It allow to save time on `import hidet`. 

Test manually:
 - 6th version of hidet stops old process and start new one. 
 - Comp time for individual job improved

### DIFF
--- a/apps/compile_server/resources/compilation.py
+++ b/apps/compile_server/resources/compilation.py
@@ -17,6 +17,20 @@ from filelock import FileLock
 
 from .compile_worker import CompilationWorkers
 
+'''
+The compilation server will launch as many Flask applications as there are vCPUs (using gunicorn). 
+Each Flask application (i.e., our compilation server process) will handle the requests, 
+and there will be at most vCPU number of requests being processed at the same time. 
+Each process will maintain a pool of compilation workers with max_workers=5 (i.e., independent processes),
+with a specific version of hidet that has been imported in every process. 
+The job will (try to) be dispatched to a worker with the same hidet version first. 
+If no such worker exists, then a new one will be created to replace an existing one.
+
+Increasing the `max_workers` in `CompilationWorkers` init will (potentially) consume more memory 
+(thanks to fork, this problem will not get severe) and create more processes (max_workers * vCPU in total). 
+Reducing the max_workers will reduce the opportunity to avoid importing hidet with the same version in nearby jobs.
+'''
+
 lock = threading.Lock()
 logger = logging.Logger(__name__)
 
@@ -161,8 +175,7 @@ class CompilationResource(Resource):
             with lock:  # Only one thread can access the following code at the same time
                 print('[{}] Start compiling: {}'.format(pid, job_id[:16]), flush=True)
                 start_time = time.time()
-                compilation_workers.submit_job(job_id, version_path)
-                compilation_workers.wait_all_jobs_finished()
+                compilation_workers.run_and_wait_job(job_id, version_path)
                 end_time = time.time()
 
             # respond to the client
--- a/apps/compile_server/resources/compilation.py
+++ b/apps/compile_server/resources/compilation.py
@@ -1,11 +1,9 @@
-from typing import Dict, Any, List, Tuple
+from typing import Dict, Any, Tuple
 import time
 import re
-import sys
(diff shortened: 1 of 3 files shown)


apps/compile_server/resources/compile_worker.py
```diff
@@ -88,12 +83,70 @@ def compile_job(job_id: str):
...
+        self.workers[version_path] = (worker, job_queue)
+        return self.workers[version_path]
+
+    def submit_job(self, job_id, version_path):
+        _, job_queue = self._get_or_create_worker(version_path)
+        job_queue.put(job_id)
+
+    def wait_all_jobs_finished(self):
+        # multiprocessing.Queue.get() waits until a new item is available
+        self.result_queue.get()
```
yaoyaoding, line 146: Since we do not plan to use these workers to run in parallel, we can merge this two functions by returning the result directly after submitting a job. 

The current `wait_all_jobs_finished()` is ambegiuous since it only tries to wait until one job is finished and return it.

apps/compile_server/resources/compilation.py
```diff
@@ -17,16 +15,18 @@
...
-repos_dir = os.path.join(os.getcwd(), 'repos')
-commits_dir = os.path.join(os.getcwd(), 'commits')
-results_dir = os.path.join(os.getcwd(), 'results')
+JOBS_DIR = os.path.join(os.getcwd(), 'jobs')
+REPOS_DIR = os.path.join(os.getcwd(), 'repos')
+COMMITS_DIR = os.path.join(os.getcwd(), 'commits')
+RESULTS_DIR = os.path.join(os.getcwd(), 'results')
 
-compile_script = os.path.join(os.path.dirname(__file__), 'compile_worker.py')
+compilation_workers = CompilationWorkers(max_workers=5)
```
yaoyaoding, line 29: Better to add a comment on how the number of 5 comes from.

The compilation server will launch as many as vCPU of Flask applications (by gunicorn). Each Flash application (i.e., our compilation server process) will handle the requests and there will be at most vCPU number of requests been processed at the same time. Each process will maintain a pool of compilation workers with `max_workers=5` number workers (i.e., independent process), with a specific version of hidet that has been imported. The job will (try to) be dispatched to a worker with the same hidet version first, if no such a worker, then a new one will be created to replace an existing one.

yaoyaoding (approved):
LGTM in general, left a few minor comments. Thanks @vadiklyutiy !

apps/compile_server/resources/compilation.py
yaoyaoding: Increase the `max_workers` will (potentially) consume more memory and create more process (`max_workers` * vCPU in total). Reduce the `max_workers` will reduce the opportunity to avoid importing hidet with the same version in near-by jobs.

apps/compile_server/resources/compile_worker.py
```diff
@@ -88,12 +83,70 @@ def compile_job(job_id: str):
...
-    exit(compile_job(args.job_id))
-
-
-if __name__ == '__main__':
-    main()
+# Worker process function to handle compilation jobs using a specific version of the 'hidet' module.
+def worker_process(version, job_queue, result_queue, parent_pid):
+    sys.path.insert(0, os.path.join(version, 'python'))  # Ensure the version path is first in sys.path
+    hidet = importlib.import_module("hidet")  # Load the specific version
+    importlib.reload(hidet)  # Reload to ensure correct version
```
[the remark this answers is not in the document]
vadiklyutiy: Right now, not necessary. 
But more safe in potential future case if somebody call import hidet inside `CompilationResource.post()`. 

If you think we don't need I can delete. Don't have strong opinion that we should leave it.

apps/compile_server/resources/compile_worker.py
vadiklyutiy: ok

apps/compile_server/resources/compilation.py
vadiklyutiy: added

PR #490, opened 2025-01-31
[FEATURE] In compilation server clean memory after every compilation 

vadiklyutiy:
In compilation server clean memory after every compilation 

Consuming memory after compilation might be around 2GB. It is not used but just allocated memory. Free it.


vadiklyutiy:
@yaoyaoding 
kindly remind about this PR

